PR #74, opened 2019-02-01
Fix build on clean sbt installation

migesok:
It's a good practice to have project buildable by just installed sbt
without additional configuration.
Without this option test target fails with OutOfMemoryError: Metaspace


t3hnar:
@migesok I'm completely agree.

PR #114, opened 2020-01-30
Metrics for conversion between events and payload

lavrov:


journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -42,7 +42,8 @@ object Journals {
     config: JournalConfig,
     origin: Option[Origin],
     eventualJournal: EventualJournal[F],
-    metrics: Option[JournalMetrics[F]],
+    journalMetrics: Option[JournalMetrics[F]],
+    conversionMetrics: Option[ConversionMetrics[F]],
```
t3hnar: why it's not part of `JournalMetrics` ?

journal/src/main/scala/com/evolutiongaming/kafka/journal/conversions/EventsToPayloadMetrics.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  def of[F[_]: Monad](
+    registry: CollectorRegistry[F],
+    prefix: String = "journal"
+  ): Resource[F, EventsToPayloadMetrics[F]] = {
+
+    val durationSummary = registry.summary(
+      name = s"${prefix}_events_to_payload_duration",
+      help = "Journal events to payload conversion duration in seconds",
+      quantiles = Quantiles(Quantile(0.9, 0.05), Quantile(0.99, 0.005)),
```
t3hnar: why don't we use the same as in other places ?

lavrov: They are the same

PR #136, opened 2020-03-30
Configurable JsonCodec in Replicator

lavrov:


journal/src/main/scala/com/evolutiongaming/kafka/journal/JsonCodec.scala
```diff
@@ -89,51 +89,77 @@ object JsonCodec {
   }
 
 
-  final case class Decode[F[_]](fromBytes: FromBytes[F, JsValue])
+  final case class Decode[F[_]](fromBytes: FromBytes[F, JsValue], fromStr: String => F[JsValue])
```
t3hnar: why do we need `fromStr` at all here?

lavrov: To cover parsing strings.

 https://github.com/evolution-gaming/kafka-journal/pull/136/files#diff-a2d895d6800c92630153ca061a462ad3R89

t3hnar: but why should we cover parsing strings at all ?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
```diff
@@ -175,7 +184,7 @@ object JournalStatements {
               val payload = payloadType.map {
                 case PayloadType.Binary => row.decode[Option[Payload.Binary]]("payload_bin") getOrElse Payload.Binary.empty
                 case PayloadType.Text   => row.decode[Payload.Text]("payload_txt")
-                case PayloadType.Json   => row.decode[Payload.Json]("payload_txt")
+                case PayloadType.Json   => row.decode[Payload.Json]("payload_txt")(decodeByNameJson)
```
t3hnar, line 187: why explicit ?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/PlayJsonHelper.scala
```diff
@@ -75,17 +81,20 @@ object PlayJsonHelper {
...
+  }
 
   
-  def decodeByNameFromReads[A](implicit reads: Reads[A]): DecodeByName[A] = DecodeByName[String].map { a =>
-    val jsValue = PlayJsonJsoniter.deserializeFromStr(a) getOrElse Json.parse(a)
-    // TODO not use `as`
-    jsValue.as(reads)
+  def decodeByNameFromReads[A](implicit reads: Reads[A], decode: JsonCodec.Decode[Try]): DecodeByName[A] = {
+    DecodeByName[String].map { str =>
+      val jsValue = decode.fromStr(str).get
```
t3hnar, line 94: although I can accept the review, could you please benchmark this against direct PlayJson from String ?

journal/src/main/scala/com/evolutiongaming/kafka/journal/JsonCodec.scala
```diff
@@ -40,51 +40,59 @@ object JsonCodec {
...
+  }
+
+
+  final case class Encode[F[_]](toBytes: ToBytes[F, JsValue])
 
   object Encode {
 
     implicit def fromCodec[F[_]](implicit codec: JsonCodec[F]): Encode[F] = codec.encode
 
+    implicit def encodeTry[F[_] : ToTry](implicit encode: Encode[F]): Encode[Try] = encode.mapK(ToTry.functionK)
```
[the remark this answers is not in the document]
lavrov: Removed

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
lavrov: Instance of `DecodeByName[Payload.Json]` is cached to avoid creation for each call. The explicit usage is for making this detail more explicit ;)

journal/src/main/scala/com/evolutiongaming/kafka/journal/JsonCodec.scala
```diff
@@ -16,7 +16,7 @@ final case class JsonCodec[F[_]](
 
 object JsonCodec {
 
-  def apply[F[_]](implicit F: JsonCodec[F]): JsonCodec[F] = F
+  def summon[F[_]](implicit F: JsonCodec[F]): JsonCodec[F] = F
```
rtar, line 19: @lavrov @t3hnar Why it is renamed to `summon` instead of `apply`? I think summoning through `apply` is universally accepted practice.

journal/src/main/scala/com/evolutiongaming/kafka/journal/Events.scala
```diff
@@ -28,7 +31,9 @@ object Events {
     codecs.choice(version1, version0, default)
   }
 
-  implicit def eventsToBytes[F[_] : FromAttempt]: ToBytes[F, Events] = ToBytes.fromEncoder
+  implicit def eventsToBytes[F[_] : FromAttempt](implicit jsonCodec: JsonCodec[Try]): ToBytes[F, Events] =
+    ToBytes.fromEncoder
 
-  implicit def eventsFromBytes[F[_] : FromAttempt]: FromBytes[F, Events] = FromBytes.fromDecoder
+  implicit def eventsFromBytes[F[_] : FromAttempt](implicit jsonCodec: JsonCodec[Try]): FromBytes[F, Events] =
```
rtar, line 37: Why whole `JsonCodec[Try]` is required here, rather than `JsonCodec.Decode[Try]`?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Events.scala
```diff
@@ -28,7 +31,9 @@ object Events {
     codecs.choice(version1, version0, default)
   }
 
-  implicit def eventsToBytes[F[_] : FromAttempt]: ToBytes[F, Events] = ToBytes.fromEncoder
+  implicit def eventsToBytes[F[_] : FromAttempt](implicit jsonCodec: JsonCodec[Try]): ToBytes[F, Events] =
```
rtar, line 34: Why whole `JsonCodec[Try]` is required here, rather than `JsonCodec.Encode[Try]`?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -86,8 +90,9 @@ object Journals {
     log: Log[F],
     conversionMetrics: Option[ConversionMetrics[F]]
   ): Journals[F] = {
-    implicit val fromAttempt = FromAttempt.lift[F]
-    implicit val fromJsResult = FromJsResult.lift[F]
+    implicit val fromAttempt: FromAttempt[F]   = FromAttempt.lift[F]
+    implicit val fromJsResult: FromJsResult[F] = FromJsResult.lift[F]
+    implicit val jsonCodec: JsonCodec[Try]     = JsonCodec.summon[F].mapK(ToTry.functionK)
```
rtar, line 95: I see a lot of `summon[F].mapK(ToTry.functionK)` calls. Was it really intended to call it manually everywhere?

t3hnar: @rtar what do you propose instead ?

journal/src/main/scala/com/evolutiongaming/kafka/journal/JsonCodec.scala
t3hnar: We agreed with @lavrov to start moving away from use of `apply` method for "summoning" implicit

- we'd better reserve `apply` as factory method
- `summon` is reserved keyword for exact this purpose in scala 3

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
```diff
@@ -83,7 +88,7 @@ object JournalStatements {
               val (text, bytes) = payload match {
                 case payload: Payload.Binary => (None, Some(payload))
                 case payload: Payload.Text   => (Some(payload.value), None)
-                case payload: Payload.Json   => (Some(PlayJsonJsoniter.serializeToStr(payload.value)), None)
+                case payload: Payload.Json   => (Some(encodeTry.toStr(payload.value).get), None)
```
rtar, line 91: Do we really call `Try.get` here or my eyes lie to me?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Events.scala
lavrov: It uses `FromBytes.fromDecoder` which requires `scodec.Decoder`. The decoder instance is provided by `scodec.Codec` which is in turn requires both `Encode` and `Decode` i.e `JsonCodec`. To fix this we would have to split codec into encoder and decoder.

journal/src/main/scala/com/evolutiongaming/kafka/journal/Events.scala
lavrov: Same: implemented in terms of `scodec.Codec`

journal/src/main/scala/com/evolutiongaming/kafka/journal/Events.scala
rtar: Do you mean `Codec[Events]`? I think it is an excellent idea to split it.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
lavrov: Yeah, this could be fixed

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
rtar: Big picture:
1. `JsonCodec[Try]` is required to construct `Codec[Events]` in `Events.codecEvents` method. It is not used there explicitly at all, but just passed into the two methods: `Event.codecEvent` and `ScodecHelper.formatCodec`.
2. It is used directly in `ScodecHelper.formatCodec`.
3. It is used implicitly in `Event.codecEvent` to call `Payload.Json.codecJson`.
4. `Payload.Json.codecJson` just calls `ScodecHelper.formatCodec` again.
5. `ScodecHelper.formatCodec` uses `Attempt.fromTry` on every value received from `JsonCodec`.

What I suggest:
1. `Events.codecEvents` etc. does not have to know anything about `Try`. It only needs `Codec[Event]` and `Codec[PayloadMetadata]`.
2. If one does not want to pass them, but construct in `Events` object instead, the only thing which is required is a function `Format[T] => Codec[T]`, or FunctionK[Format, Codec], no `Try` is involved or required.
3. `JsonCodec` may, indeed, provide implementation of that function using `Try`, but, I guess, it belongs somewhere near `JsonCodec`, i.e. in companion object.

journal/src/main/scala/com/evolutiongaming/kafka/journal/JsonCodec.scala
rtar: I have googled a bit, and I could not find any Scala 2 project that uses `summon` in such a way. It means, if we go that way, we will be pioneers and will confuse every single new developer a bit. Not in a critical way, `summon` is quite a clear and well chosen name, but there will be a long period of time when both `apply` and `summon` are used in our projects. Are we sure we want to be pioneers in that area?

> we'd better reserve apply as factory method
I think we are not breaking the commonplace `apply` contract anyhow when using it for summoning (we are returning instance of the class after all), and we can still use `apply` for factory methods even if we used it for summoning, I think, or not?

t3hnar: > Are we sure we want to be pioneers in that area?

I personally sure, using `apply` for that have some drawbacks

rtar: @t3hnar I agree about drawbacks, but it just came to be that we might be breaking the expectations of most typeclass libraries, including, but not limited to:
- https://typelevel.org/cats-tagless/
- https://github.com/typelevel/simulacrum

I think we need to be careful here. I see that some libraries do not require `apply`, i.e. this one, so it might still be okay, but I am a bit worried:
https://github.com/augustjune/context-applied

PR #149, opened 2020-04-18
Not measure already replicated records twice to avoid ruin the metrics

olegkovalenko:

### DIFF
--- a/journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedKeyJournal.scala
+++ b/journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedKeyJournal.scala
@@ -7,6 +7,7 @@ import cats.implicits._
 import cats.{Applicative, FlatMap, ~>}
 import com.evolutiongaming.catshelper.{ApplicativeThrowable, Log}
 import com.evolutiongaming.kafka.journal._
+import com.evolutiongaming.kafka.journal.eventual.ReplicatedKeyJournal.Changed
 import com.evolutiongaming.skafka.{Offset, Topic}
 import com.evolutiongaming.smetrics._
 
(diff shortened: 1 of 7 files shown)


journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedKeyJournal.scala
```diff
@@ -18,19 +18,19 @@ trait ReplicatedKeyJournal[F[_]] {
     timestamp: Instant,
     expireAfter: Option[ExpireAfter],
     events: Nel[EventRecord]
-  ): F[Unit]
+  ): F[Boolean]
```
t3hnar, line 21: can we replace `Boolean` everywhere with type alias `Updated` or `Changed` ?

journal/src/test/scala/com/evolutiongaming/kafka/journal/eventual/EventualJournalSpec.scala
```diff
@@ -507,11 +507,11 @@ object EventualJournalSpec {
 
         def append(partitionOffset: PartitionOffset, events: Nel[EventRecord]) = {
           // TODO expiry: define expireAfter and test
-          journal.append(key, partitionOffset, timestamp, none, events)
+          journal.append(key, partitionOffset, timestamp, none, events).map(_ => ())
```
t3hnar, line 510: .void :)

journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedKeyJournal.scala
olegkovalenko: good point

journal/src/test/scala/com/evolutiongaming/kafka/journal/eventual/EventualJournalSpec.scala
olegkovalenko: nice, didn't know about this one :)

olegkovalenko:
Thank you!

PR #341, opened 2022-01-18
WIP: Ce3 migration

Z1kkurat:
This is a very rough draft. Current issues:
- some tests including running Cassandra don't work on my machine due to 3.x Cassandra driver not working nicely with Java module system (getting `Cause: java.lang.reflect.InaccessibleObjectException: Unable to make field private int java.io.FileDescriptor.fd accessible: module java.base does not "opens java.io" to unnamed module @38899305`) and the absence of native libraries for M1 Silicon
- the code itself is a mess, it needs polishing
- not all unit tests pass, I'm addressing them one by one


t3hnar:
@Z1kkurat thanks for doing this

> Cause: java.lang.reflect.InaccessibleObjectException: Unable to make field private int java.io.FileDescriptor.fd accessible: module java.base does not "opens java.io" to unnamed module @38899305

should be possible to temporary work around with adding to `.sbtopts` 
```
-J--add-opens=java.base/ java.io=ALL-UNNAMED
```

Regarding MR, should I already start reviewing?

Z1kkurat:
@t3hnar Yes, please do. I would like to get some feedback as there's a lot of changes. I'm still working on some failing tests

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraSpec.scala
```diff
@@ -34,7 +34,7 @@ class EventualCassandraSpec extends EventualJournalSpec {
           val journals = journalsOf(segmentSize, delete, segmentsFirst = segmentsFirst, segmentsSecond = segmentsSecond)
           val (_, result) = test(journals)
             .run(State.empty)
-            .get
+            .unsafeRunSync()
```
t3hnar, line 37: overall I'd prefer to not use `unsafeRunSync` without timeouts or even would better go completely `async`

t3hnar (changes requested):

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraSpec.scala
```diff
@@ -47,20 +47,24 @@ object EventualCassandraSpec {
   val selectJournalHead: MetaJournalStatements.SelectJournalHead[StateT] = {
     (key, _) => {
       StateT { state =>
-        val metaJournal = state.metaJournal.get(key)
-        (state, metaJournal)
+        IO.delay {
```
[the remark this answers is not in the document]
Z1kkurat: I changed `StateT`s from `Id` and `Try` to `IO` as it's quite hard to write `Temporal`/`Concurrent`/`Async` instances for both `Id`/`Try` (the instances are unlawful at the least, some things just can't be expressed), now even more. I think I can move this `delay` to the body of `apply` to make it more concise

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraSpec.scala
Z1kkurat: Noted, I will try to move them to async when the initial work is done and tests pass at least

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraSpec.scala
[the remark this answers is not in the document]
t3hnar: please bear in mind that kafka journal runs thousands of tests within seconds :)

would be nice to not have dramatic degradation

Z1kkurat:
Oh wow tests are green, I didn't expect that honestly

t3hnar:
> Oh wow tests are green, I didn't expect that to be honest

if you don't mess up with business logic - it should be otherwise very easy to bump cats-effect :)

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournal.scala
```diff
@@ -27,9 +28,19 @@ class KafkaJournal(config: Config) extends AsyncWriteJournal { actor =>
 
   implicit val system       : ActorSystem              = context.system
   implicit val executor     : ExecutionContextExecutor = context.dispatcher
-  implicit val contextShift : ContextShift[IO]         = IO.contextShift(executor)
-  implicit val parallel     : Parallel[IO]             = IO.ioParallel(contextShift)
-  implicit val timer        : Timer[IO]                = IO.timer(executor)
+
+  private val (blocking, blockingShutdown)   = IORuntime.createDefaultBlockingExecutionContext()
+  private val (scheduler, schedulerShutdown) = IORuntime.createDefaultScheduler()
```
t3hnar, line 33: I'd prefer reusing thread pools, could you please at least create separate task about reusing scheduler from Akka.

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/TopicReplicatorSpec.scala
```diff
@@ -57,7 +59,7 @@ class TopicReplicatorSpec extends AnyWordSpec with Matchers {
 
       val data = State(records = List(records))
 
-      val (result, _) = topicReplicator.run(data).get
+      val (result, _) = topicReplicator.run(data).unsafeRunSync()
```
t3hnar, line 62: Do you have plans on improving `unsafeRunSync()` part?

t3hnar (changes requested):
Overall nice job! Thanks.

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/TopicReplicatorSpec.scala
Z1kkurat: I switched to async suite for `ConsumeTopicTest` and `TopicReplicatorSpec`. However, there are yet `EventualCassandraSpec`, `EventualCassandraTest` and `JournalSpec` using `unsafeRunSync()` but they have quite a complex code to be easily switched to return `Future[Assertion]` as scalatest requires. I would suggest leaving them outside of the scope of this PR

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournal.scala
Z1kkurat: The main compute pool is reused here (`executor`). As for the blocking pool (`Blocker`), I haven't seen any custom definitions of that, which means it was using the default cached one (`BlockerPlatform#apply`) in which case it's no different from what we use here.
As for the scheduler, it works a bit differently in CE3. Previously, `IO.timer(ec)` was used where `ec` was the thread pool to execute tasks after schedule and actual scheduling was performed on a global `ScheduledExecutorService` (`IOTimer#apply`). So we didn't specify a custom `ScheduledExecutorService`, only the thread pool for binding continuations (`ec`). Now we don't need this `ec` to bind continuations and the scheduler itself is still a default one (as before).

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1326,29 +1327,27 @@ object ReplicatedCassandraTest {
...
+    override def canceled: StateT[Unit] = ().pure[StateT]
+
+    override def onCancel[A](fa: StateT[A], fin: StateT[Unit]): StateT[A] = fa
+
+    override def suspend[A](hint: Sync.Type)(thunk: => A): StateT[A] = cats.data.StateT.pure(thunk)
+
+    override def monotonic: StateT[FiniteDuration] = ???
 
-    def suspend[A](thunk: => StateT[A]) = thunk
+    override def realTime: StateT[FiniteDuration] = ???
```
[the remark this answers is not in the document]
Z1kkurat: Done, faked it plus removed an unused `TestSync` file

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournal.scala
t3hnar: > As for the blocking pool (Blocker), I haven't seen any custom definitions of that, which means it was using the default cached one (BlockerPlatform#apply) in which case it's no different from what we use here.

Afaik we used custom resizable pool 

```
Executors.blocking
```

Could we keep it used ?

Z1kkurat: I couldn't find anything in the current `master` branch by searching for `Blocker` and `Executors.blocking` is used only in a couple of places (and `KafkaJournal` isn't one of them [link](https://github.com/evolution-gaming/kafka-journal/search?q=Executors.blocking)) and I haven't changed that, so I'm not sure what you mean.

I can use `Executors.blocking` here, but there's a catch - it returns `Resource` which we need to allocate and for that we need `IORuntime` which is basically created with the result of that Resource 😄 
I can do something like this here, if it's OK with you:
```scala
  val (blocking, blockingShutdown) = 
    Executors.blocking[IO]("kafka-journal-blocking")
      .allocated
      .unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
  implicit val ioRuntime: IORuntime = IORuntime(
    compute = executor,
    blocking = blocking,
    scheduler = scheduler,
    shutdown = () => {
      blockingShutdown.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
      schedulerShutdown()
    },
    config = IORuntimeConfig()
  )
```

However, that's not very clean and I'm not sure if there's much of a difference, considering that it has the same implementation as `IORuntime.createDefaultBlockingExecutionContext()`:

Source:
[IORuntime.createDefaultBlockingExecutionContext](https://github.com/typelevel/cats-effect/blob/27307cfe6cd9ea3cb230a8f9be5e16f624a914cd/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala#L106-L116) where `Executors.newCachedThreadPool` has the following implementation:
```java
public static ExecutorService newCachedThreadPool(ThreadFactory threadFactory) {
        return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                      60L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>(),
                                      threadFactory);
    }
```

[Executors.blocking](https://github.com/evolution-gaming/kafka-journal/blob/14b9916782ab05a3e9da1199aac4958f65e5409e/journal/src/main/scala/com/evolutiongaming/kafka/journal/util/Executors.scala#L14-L23) where `ThreadPoolOf[F](2, Int.MaxValue, threadFactory)` does just [the same](https://github.com/evolution-gaming/kafka-journal/blob/14b9916782ab05a3e9da1199aac4958f65e5409e/journal/src/main/scala/com/evolutiongaming/kafka/journal/execution/ThreadPoolOf.scala) (except for `minSize` which is 2 instead of 0 and also CE version spawns daemonic threads) and it's also not configurable in terms of size, you can only pass a thread prefix, just like you can pass it  to `IORuntime.createDefaultBlockingExecutionContext`.

Z1kkurat: I believe it's a common approach in cats-effect to use an unbounded cached pool for blocking operations. Even in cats-effect 2.x.x, if we take a look at the implementation which is summoned when using `Blocker[IO]` it uses the same unbounded `Executors.newCachedThreadPool` [link](https://github.com/typelevel/cats-effect/blob/series/2.x/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala#L31-L38)

t3hnar: I'm happy to see things evolving this way!

It used to be a different opinion of community, code in subject existed before `Blocker`… and even before we started to adopt cats & cats-effect :)

Z1kkurat:
Just in case -- I forked [series/0.0.x](https://github.com/evolution-gaming/kafka-journal/tree/series/0.0.x) branch from `master` and pushed it to be able to back-port any changes easily

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournal.scala
```diff
@@ -27,9 +28,19 @@ class KafkaJournal(config: Config) extends AsyncWriteJournal { actor =>
 
   implicit val system       : ActorSystem              = context.system
   implicit val executor     : ExecutionContextExecutor = context.dispatcher
-  implicit val contextShift : ContextShift[IO]         = IO.contextShift(executor)
-  implicit val parallel     : Parallel[IO]             = IO.ioParallel(contextShift)
-  implicit val timer        : Timer[IO]                = IO.timer(executor)
+
+  private val (blocking, blockingShutdown)   = IORuntime.createDefaultBlockingExecutionContext("kafka-journal-blocking")
```
t3hnar, line 32: I checked internals, I think this is fine to use consistently this one instead of custom `Executors.blocking`

t3hnar (changes requested):

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournal.scala
Z1kkurat: Good, thanks! As I've already mentioned, the only differences I've spotted is minSize = 0 instead of 2 and daemonic threads instead of non-daemonic

Z1kkurat:
@t3hnar Is there anything else you would like me to address here?

t3hnar:
> @t3hnar Is there anything else you would like me to address here?

@Z1kkurat Yep, replace usage of `Executors.blocking` with `IORuntime.createDefaultBlockingExecutionContext`

Z1kkurat:
> @Z1kkurat Yep, replace usage of `Executors.blocking` with `IORuntime.createDefaultBlockingExecutionContext`

Got it, will do. 

There's a slight catch: `IORuntime.createDefaultBlockingExecutionContext` returns `(ExecutionContext, () => Unit)` and `Executors.blocking` returns `Resource[F, ExecutionContextExecutorService]`.

I think we have a couple of options about that:
-  we could wrap these usages of `IORuntime.createDefaultBlockingExecutionContext` into `Resource` manually at call-sites. It's not a big deal as it's used only in 3 places in this codebase: `JournalAdapter`, `AppendReplicateApp` and `IntegrationSuite`.
- another option would be to introduce a new method in `Executors` that will do this wrapping and replace usages of `Executors.blocking` with that.

Both options have the same disadvantage, however: `IORuntime` is an implementation detail of CE's runtime (it's in the `cats-effect` library which contains the implementation of `IO` and not in `cats-effect-kernel` which contains abstractions) so, ideally, it should leak neither to `Executors` which is polymorphic by `F` nor to these usage sites which are also polymorphic by `F`

Z1kkurat:
@t3hnar as discussed, I removed blocking pool creation from `JournalAdapter` and a couple of other related places including tests. Now there are no more usages of `Executors.blocking` in the codebase

t3hnar (approved):

PR #366, opened 2022-05-03
Remove redundant `prometheus` dependency

odbc:
Remove redundant `prometheus` dependency


persistence/src/main/resources/reference.conf
```diff
@@ -32,6 +32,7 @@ evolutiongaming.kafka-journal.persistence {
         max-poll-records = 1000
         auto-offset-reset = "earliest"
         auto-commit = false
+        client-rack = ${?KAFKA_CLIENT_RACK}
```
t3hnar, line 35: this is an open sourced library, we should not leak something like `KAFKA_CLIENT_RACK` here

PR #479, opened 2023-06-23
Remove dependency to com.evolutiongaming:scache

Z1kkurat:


Z1kkurat:
`cassandra-all` 3.x (that we use for testing via `cassandra-launcher`) depends on an older logback version that renamed one of the classes used directly in Cassandra. That's why tests fail with `java.lang.NoClassDefFoundError: ch/qos/logback/core/hook/DelayingShutdownHook`. We'd need to update `cassandra-all` in `cassandra-launcher` to something more up-to-date, hoping that it would use a later version of logback. But newer Cassandra 4.x doesn't run on Java 11+ without additional VM options (such as `--add-exports`) due to heavy reflection usage. Because of this, updating `cassandra-helper` could prove quite difficult as anyone using that library would have to add these VM options. The error when using `cassandra-all` 4.x in `cassandra-launcher`:
```
15:47:26.816 ERROR o.a.c.i.u.FileUtils FATAL: Cassandra is unable to access required classes. This usually means it has been run without the aid of the standard startup scripts or the scripts have been edited. If this was intentional, and you are attempting to use Java 11+ you may need to add the --add-exports and --add-opens jvm options from either jvm11-server.options or jvm11-client.options
java.lang.IllegalAccessException: access to public member failed: jdk.internal.ref.Cleaner.clean[Ljava.lang.Object;@620f9ce5/invokeVirtual, from org.apache.cassandra.io.util.FileUtils (unnamed module @72ce4d1d)

```

edubrovski (approved):

t3hnar (approved):

PR #500, opened 2023-07-21
Create a pool of kafka consumers

edubrovski:
Currently a KafkaConsumer is created for each actor (not sure, need to check) on recovery. This can lead to creation of thousands of Kafka consumers which will crush the app with OOM when they fill their buffers with messages.

The idea of this PR is to limit the number of Kafka consumers. I think fixed size pool is the best solution because it's the easiest to implement.

Pool size configuration is optional, None by default. In the future versions we'll deprecate the None variant and make consumer pool mandatory, maybe we'll set autoscaling with number of CPUs pool configuration as default.


journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumerPool.scala
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * A fixed size pool.
+ *
+ * Semantically blocks on `acquire` if there are no consumers available.
+ *
+ * Make sure to release all consumers before shutting the pool down.
+ */
+private[journal] trait ConsumerPool[F[_]] {
+
+  def acquire: F[Consumer[F]]
```
t3hnar, line 20: why do you need such api? can't we just keep using `Resource[Consumer]` ?

t3hnar (changes requested):

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumerPool.scala
edubrovski: Yeah, good point, I should probably use Resource instead

migesok:
Another important thing: in the end, waiting for a pooled resource should be limited to avoid causing OOMs. Either with a fixed size queue or a IO-cancelling timeout here or somewhere higher up the chain of calls.

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumerPool.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def release(consumer: Consumer[F]): F[Unit]
+}
+
+private[journal] object ConsumerPool {
+
+  case class ConsumerRef[F[_]](consumer: Consumer[F], deallocate: F[Unit])
```
migesok, line 27: I don't see consumer reset (unassign partitions) before returning to pool factored in.

edubrovski: I read the javadoc on `org.apache.KafkaConsumer.assign` and it says
```
Manually assign a list of partitions to this consumer. This interface does not allow for incremental assignment and will replace the previous assignment (if there is one).
```

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/Pool.scala
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * `acquire` function of the Resource semantically blocks if there are no objects available.
+   *
+   * `release` function of the Resource returns the object to the pool.
+   * Note that it will NOT deallocate the object.
+   */
+  def borrow: Resource[F, O]
+}
+
+private[journal] object Pool {
```
migesok, line 20: I'd suggest to call it "ResourcePool", a bit more descriptive.

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumerPool.scala
migesok: > what can happen if we don't unassign partitions before returning to the pool

The records buffered by the consumer will be kept in memory until the next use, the consumer might continue some background work of fetching records behind the scene even if we don't use it anymore. Not sure if there could be any other consequences. The unassign btw might be done on different level - the stream outside of the pool might do it on finalization.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/Pool.scala
```diff
@@ -0,0 +1,77 @@
...
+    implicit F: GenConcurrent[F, Throwable],
+  ) extends Pool[F, O] {
+
+    override def borrow: Resource[F, O] =
+      Resource.make(acquire)(release).map(_.obj)
+
+    def acquire: F[ObjectRef[F, O]] =
+      for {
+        ref <- unused.take
+        _ <- used.update(_ + ref)
```
migesok, line 58: I'd suggest in order to guarantee that it works 100% to use identity equals and hashcode (default object ones) in ObjectRef. Otherwise it depends on identity of objects inside it (that's how case class equals and hashcode works). The easiest way to do it should be making ObjectRef just a plain final class with `val`s. Don't forget to leave a comment there that it's a plain class to leverage the identity equality of Java Object and use it in sets.

edubrovski: > Otherwise it depends on identity of objects inside it
Yes, why do you think it's bad? User code always returns the same object (pointing to the same address in memory) that it got from the pool.

edubrovski: I guess this is not relevant anymore since I changed pool implementation to the one you proposed?

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumeActionRecords.scala
```diff
@@ -18,14 +22,18 @@ trait ConsumeActionRecords[F[_]] {
...
-    consumer: Resource[F, Journals.Consumer[F]],
-    log: Log[F])(implicit
-    consRecordToActionRecord: ConsRecordToActionRecord[F]
+  def apply[F[_] : BracketThrowable: Clock](
+    consumerPool: ResourcePool[F, Journals.Consumer[F]],
+    log: Log[F],
+    poolMetrics: ConsumerPoolMetrics[F],
+  )(implicit
+    consRecordToActionRecord: ConsRecordToActionRecord[F],
+    genConcurrent: GenConcurrent[F, Throwable]
```
rtar, line 31: I wonder why such parameter is required? Why not just `Concurrent[F]` instead of `GenConcurrent[F, Throwable]`?

Also, why named parameter is required, would not it make sense to just replace `F[_] : BracketThrowable` with `F[_]: Concurrent`? :thinking:

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -92,22 +95,25 @@ object Journals {
     consumer: Resource[F, Consumer[F]],
     eventualJournal: EventualJournal[F],
     headCache: HeadCache[F],
+    consumerPoolSize: Int,
+    consumerPoolMetrics: Option[ConsumerPoolMetrics[F]],
     log: Log[F],
     conversionMetrics: Option[ConversionMetrics[F]]
-  )(implicit F: MonadCancel[F, Throwable]): Journals[F] = {
+  )(implicit F: GenConcurrent[F, Throwable]): Resource[F, Journals[F]] = {
```
rtar, line 102: As this method now returns `Resource[F, Journals[F]]` instead of `Journals[F]` it makes sense to rename it from `apply` to `make` to avoid confusing the library users.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,39 @@
...
+   */
+  def borrow: Resource[F, O]
+}
+
+private[journal] object ResourcePool {
+
+  /**
+   *  Objects are allocated when a pool is created and deallocated when it shuts down.
+   */
+  def fixedSize[F[_], O](objectOf: Resource[F, O], poolSize: Int, log: Log[F])(
```
rtar, line 25: I do not see `log` parameter being used in the code :thinking:

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumeActionRecords.scala
rtar: I see you fixed this in `ConsumeActionRecords.of`. May be the same fix could also be done in `ResourcePool.fixedSize`?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
rtar: I see you changed it to `of` now. I am fine to keep it `of`, though I have a (false?) memory that the usual style is the following in case you have some class such as `MyClass`:
1. Use `apply` if method returns `MyClass`.
2. Use `of` if method returns `F[MyClass]`.
3. Use `make` if method returns `Resource[F, MyClass]`.

I am not sure about it, so it is okay to keep `of`.

edubrovski: I don't have a preference for `of`/`make`. Let's change it to `make` if this is the style we follow. I called it `of` because we already have an `of` method in `Journals` that returns a `Resource` and I thought it would be weird that one method is called `of` and another is called `make`

rtar:
I have accidentally stumbled upon this structure recently. If only maximum of consumers is required, not the reuse of consumers themselves, may be it could be considered to be put in use? I understand that pooling/reuse may make everything faster, but it may also cause interesting bugs such as broken consumers.

https://typelevel.org/cats-effect/docs/std/backpressure

edubrovski:
@rtar , I see, this might be a good idea. It's true that we only need to limit the maximum number of consumers. I'll look into this Backpressure

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumerPool.scala
edubrovski: I agree that this might be a good idea, but looks like it will lead to lots of code changes. Can we maybe do it in a separate MR?

edubrovski:
https://github.com/evolution-gaming/cats-helper/pull/237 will allow us to simplify `consumerWithMetrics`. Right now `consumerWithMetrics` is pretty weird because I had to use `Deferred` (alternatively I could use `allocated`).

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,36 @@
...
+   */
+  def borrow: Resource[F, O]
+}
+
+private[journal] object ResourcePool {
+
+  /**
+   *  Objects are allocated when a pool is created and deallocated when it shuts down.
+   */
+  def fixedSize[F[_]: Concurrent, O](makeObject: Resource[F, O], poolSize: Int): Resource[F, ResourcePool[F, O]] = {
```
t3hnar, line 24: makeObject  -> resource?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,36 @@
...
+    for {
+      objects <- Vector.fill(poolSize)(makeObject).sequence
```
t3hnar, line 26: should it be `lazy`? do we want to keep the pool `full` even if we never need it?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
```diff
@@ -329,4 +330,34 @@ object Journal {
 
     implicit val configReaderCallTimeThresholds: ConfigReader[CallTimeThresholds] = deriveReader[CallTimeThresholds]
   }
+
+  sealed trait ConsumerPoolSize {
+
+    def calculate[F[_]: Runtime: Applicative]: F[Int] =
+      this match {
+        case ConsumerPoolSize.Fixed(size) =>
```
t3hnar, line 338: nice idea, but… do we ever need this?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -85,7 +86,56 @@ object Journals {
...
+        log,
+        conversionMetrics
+      )
+    } yield {
+      val withLog = journal.withLog(log, callTimeThresholds)
+      journalMetrics.fold(withLog) { metrics => withLog.withMetrics(metrics) }
+    }
+  }
 
+  @deprecated("use `make`", "2023-07-26")
```
t3hnar, line 138: let's be consistent with naming, let's use `apply1`

journal/src/main/scala/com/evolutiongaming/kafka/journal/RecoveryConsumerPoolMetrics.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+trait RecoveryConsumerPoolMetrics[F[_]] {
+
+  def acquireTime(topic: Topic, time: FiniteDuration): F[Unit]
+
+  def useTime(topic: Topic, time: FiniteDuration): F[Unit]
+}
+
+object RecoveryConsumerPoolMetrics {
```
t3hnar, line 19: why do we need `Recovery` keyword here?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -107,6 +157,41 @@ object Journals {
...
+     log: Log[F],
+     conversionMetrics: Option[ConversionMetrics[F]]
+  ): Resource[F, Journals[F]] = {
+    implicit val fromAttempt: FromAttempt[F]   = FromAttempt.lift[F]
+    implicit val fromJsResult: FromJsResult[F] = FromJsResult.lift[F]
+
+    for {
+      poolSize    <- consumerPoolSize.calculate.toResource
+      _           <- log.debug(s"Creating a consumer pool of size $poolSize").toResource
+      pool        <- ResourcePool.fixedSize[F, Consumer[F]](consumer, poolSize)
```
t3hnar, line 184: not sure you safely can use `consumer` here, you might need to add more functionality to `release`, please check whether you need to `reset` consumer's state?

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumeActionRecords.scala
```diff
@@ -18,10 +22,44 @@ trait ConsumeActionRecords[F[_]] {
...
 
+  @deprecated("use `of` instead", "2023-07-26")
   def apply[F[_] : BracketThrowable](
     consumer: Resource[F, Journals.Consumer[F]],
     log: Log[F])(implicit
     consRecordToActionRecord: ConsRecordToActionRecord[F]
+  ): ConsumeActionRecords[F] = apply1(_ => consumer, log)
+
+  def of[F[_] : Concurrent : Clock](
+    consumerPool: ResourcePool[F, Journals.Consumer[F]],
```
t3hnar, line 33: why not use `Resource[F, Journals.Consumer[F]]` ?

t3hnar: why this class need to know anything about underlying pool and even combine metrics? 
I think you should create separate class that takes `metrics/consumer/pool settings` and returns `Resource[consumer]`

t3hnar (changes requested):

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumeActionRecords.scala
edubrovski: >  create separate class that takes metrics/consumer/pool settings and returns Resource[consumer]
This would be nice, but the problem is I want `Topic` for writing metrics and `Topic` comes from `Key`, so I don't know `Topic` in advance when constructing this pooled consumer. Or I have to have a complicated signature like
`def pooledConsumer[F[_]](...): Resource[F, Topic => Resource[F, Consumer[F]]]`

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+    }
+  }
+
+  case class State[F[_], O](
+    objects: Map[ObjectId, AllocatedObject[F, O]],
+    waiters: Map[WaiterId, Waiter[F, O]],
+  )
+
+  def fixedSize[F[_]: Async, O](
+    resource: Resource[F, O],
```
rtar, line 50: Not insisting, but I would consider:
1. Making it `=> Resource[F, O]` instead of `Resource[F, O]` as it will follow a least surprise principle and will allow one to add `println` into this block while debugging :grin: 
2. I would consider using two parameter lists, with `resource` being a last parameter to allow syntax like following:
```scala
ResourcePool.fixedSize(my,config,parameters) { creating resource here }
```

It is not a public API, so you are welcome to ignore this comment. I just wanted to share :blush:

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+
+      def checkForAcquireTimeout(
+        waiterId: WaiterId,
+        promise: Waiter[F, O]
+      ): F[Unit] = {
+        promise.tryGet.flatMap {
+          case Some(_) =>
+            Async[F].unit
+          case None =>
+            mutex.lock.surround {
```
rtar, line 93: I see that `checkForAcquireTimeout` is called from `registerWaiter`, which is called by `acquirePromise`, which already calls `mutex.lock.surround`.

I am actually a bit surprised it does not deadlock in this case as `Mutex` is not reentrant according to documentation.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+                    for {
+                      o <- allocateObject(state)
+                      _ <- promise.complete(o.asRight)
+                    } yield ()
+                  } else {
+                    registerWaiter(state, promise)
+                  }
+              }
+            }
+          }.as(promise)
```
rtar, line 148: I think you can also use `flatTap`, if you want to avoid `as(promise)` like this:
```scala
Deferred[F, Either[Throwable, (ObjectId, O)]].flatTap { promise =>
  ... your code ...
}
```

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+                      _ <- stateRef.update { state =>
+                        state.copy(waiters = state.waiters - waiterId)
+                      }
+                      _ <- markObjectAsUsed(id, obj)
+                    } yield ()
+
+                  case None =>
+                    for {
+                      releasedAt <- Async[F].monotonic
+                      _ <- (Async[F].sleep(resourceTTL) *> checkIfExpired(id, releasedAt)).start
```
rtar, line 208: This will produce a memory leak if one and the same resource is acquired / released a lot. I.e. each release will use a bit of memory, which will only be released when `resourceTTL` is reached. I am not sure it is important, but still decided to mention it.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+    acquireTimeout: FiniteDuration,
+    resourceTTL: FiniteDuration,
+    log: Log[F],
+  ): Resource[F, ResourcePool[F, O]] = {
+
+    type StateRef = Ref[F, State[F, O]]
+
+    def poolImpl(mutex: Mutex[F], stateRef: StateRef): ResourcePool[F, O] = {
+
+      def findFreeObject(state: State[F, O]): Option[(ObjectId, AllocatedObject[F, O])] = {
```
rtar, line 61: I would consider moving this method directly to `State` case class

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+        state.objects.collectFirst {
+          case (id, obj) if obj.state.isFree => (id, obj)
+        }
+      }
+
+      def markObjectAsUsed(id: ObjectId, obj: AllocatedObject[F, O]): F[Unit] = {
+        stateRef.update { state =>
+          state.copy(objects = state.objects + (id -> obj.copy(state = AllocatedObject.State.Used)))
```
rtar, line 69: I would consider moving pure part of this method (i.e. `state.copy` part) directly to State case class...

This would allow to keep modification of `stateRef` closer to mutex allocation making it a bit more clear for a reader, potentially.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+            case Right(x) => x.pure[F]
+            case Left(err) => Async[F].raiseError(err)
+          }
+        }
+      }
+
+      def checkIfExpired(id: ObjectId, timeoutScheduledAt: FiniteDuration): F[Unit] = {
+        mutex.lock.surround {
+          stateRef.get.map(_.objects.get(id)).flatMap {
+            case Some(AllocatedObject(_, deallocate, AllocatedObject.State.Free(releasedAt))) =>
```
rtar, line 163: As `checkIfEpired` is run async, then it is completely possible to arrive here while state is `AllocatedObject.State.Used` causing the incorrect error to be reported.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+            }
+          }.as(promise)
+        }
+
+      val acquireObject: F[(ObjectId, O)] = {
+        acquirePromise.flatMap { promise =>
+          promise.get.flatMap {
+            case Right(x) => x.pure[F]
+            case Left(err) => Async[F].raiseError(err)
+          }
```
rtar, line 156: I think this block could be replaced by following:
```scala
promise.get.rethrow
```

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I see, good point. What do you propose? Periodically check for expired resources across the entire pool instead of scheduling a check every time?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: Thanks for the comments.
1. In what situations do you think `Resource[F, O]` instead of `=> Resource[F, O]` can be surprising?
2. Do you use `println` debugging? I don't use `println` even for local debugging in unit tests because then I have to think about how it plays together with cats-effect, so it's easier for me to do `log.debug` or `Sync[F].delay(println(...))`.

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
edubrovski: Thanks for the comment, will check

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,36 @@
...
+
+private[journal] object ResourcePool {
+
+  /**
+   *  Objects are allocated when a pool is created and deallocated when it shuts down.
+   */
+  def fixedSize[F[_]: Concurrent, O](objectOf: Resource[F, O], poolSize: Int): Resource[F, ResourcePool[F, O]] = {
+    for {
+      objects <- Vector.fill(poolSize)(objectOf).sequence
+      unused <- Queue.bounded[F, O](poolSize).toResource
```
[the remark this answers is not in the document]
edubrovski: We don't. Good catch! I rewrote the pool with Ref, Deferred and collections from Scala standard library.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I've written a lazy pool

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
edubrovski: Removed "fixed" variant, now we only have "scale with CPUs" variant.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,36 @@
...
+   */
+  def borrow: Resource[F, O]
+}
+
+private[journal] object ResourcePool {
+
+  /**
+   *  Objects are allocated when a pool is created and deallocated when it shuts down.
+   */
+  def fixedSize[F[_]: Concurrent, O](objectOf: Resource[F, O], poolSize: Int): Resource[F, ResourcePool[F, O]] = {
```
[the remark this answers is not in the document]
edubrovski: done

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: done

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
edubrovski: I called it `make` because @rtar wrote somewhere in this MR that we follow this convention:

- Use `apply` if method returns MyClass.
- Use `of` if method returns F[MyClass].
- Use `make` if method returns Resource[F, MyClass].

Please come to an agreement 😄

journal/src/main/scala/com/evolutiongaming/kafka/journal/RecoveryConsumerPoolMetrics.scala
edubrovski: renamed

journal/src/main/scala/com/evolutiongaming/kafka/journal/ConsumeActionRecords.scala
edubrovski: I guess it's ok if we don't add topic to consumer pool metrics. Created `PooledConsumer` object.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I don't think this can be a bottleneck. How many releases of the same resource should happen for this code to cause an OOM? 50,000? I tested locally with ResourcePoolSpec by changing `timesToAcquire` parameter. My pool did fine with 50,000 acquires. Even for 500,000 acquires there was no OOM, it was just slow.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: done

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: fixed

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+    def removeWaiter(id: WaiterId): State[F, O] =
+      State(
+        objects,
+        waiters - id
+      )
+  }
+
+  def fixedSize[F[_]: Async, O](
+    poolSize: Int,
+    acquireTimeout: FiniteDuration,
```
t3hnar, line 92: why do we need this?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+      )
+
+    def removeWaiter(id: WaiterId): State[F, O] =
+      State(
+        objects,
+        waiters - id
+      )
+  }
+
+  def fixedSize[F[_]: Async, O](
```
t3hnar, line 90: not sure you can call it a `fixedSize` now

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+              }
+            }
+        }
+      }
+
+      val acquireWaiter: F[Waiter[F, O]] =
+        Deferred[F, Either[Throwable, (ObjectId, O)]].flatTap { waiter =>
+          mutex.lock.surround {
+            stateRef.get.flatMap { state =>
+              state.findFreeObject match {
```
t3hnar, line 142: I propose that we check `state.objects.size < poolSize` and only there after try to `findFreeObject`

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+
+    object State {
+      case object Used extends State(false)
+      case class Free(releasedAt: FiniteDuration) extends State(true)
+    }
+  }
+
+  case class State[F[_], O](
+    objects: Map[ObjectId, AllocatedObject[F, O]],
+    waiters: Map[WaiterId, Waiter[F, O]],
```
t3hnar, line 45: why it's a `Map` and not a `Queue`? why do we need `WaiterId` ?

t3hnar (changes requested):

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I think this is still fine 😄 
Screenshot 2023-09-04 at 11 24 47

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+        objects,
+        waiters - id
+      )
+  }
+
+  def fixedSize[F[_]: Async, O](
+    poolSize: Int,
+    acquireTimeout: FiniteDuration,
+    idleTimeout: FiniteDuration,
+    log: Log[F],
```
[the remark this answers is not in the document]
edubrovski: 1. I use `log.error` in a couple of places.
2. This api is private to the library. Why is it so important what we pass here?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I guess what you're suggesting is to implement timeout on client side and handle fiber cancelation on pool side?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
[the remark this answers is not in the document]
t3hnar: I checked those cases of using `log.error` and I don't understand why we have them :)

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+          state <- state.get
+          _ <- notifyWaiters(state)
+          _ <- deallocateObjects(state)
+        } yield ()
+      }
+    }
+
+    val res = Resource.make {
+      for {
+        mutex <- Mutex[F]
```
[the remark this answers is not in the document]
t3hnar: I think that mutex-free implementation will be the one that is almost able to support cancellation :)

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
t3hnar: my point here is that we already have some naming patterns here:
* `apply: MyClass`
* `of: MyClass`

I propose to keep it…

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
t3hnar: I'd agree with @rtar on this one. 
I'd also prefer periodic background job that will expire un-used resources.


Imho that's too much of overhead to schedule timer per `access`

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+    }
+  }
+
+  case class State[F[_], O](
+    objects: Map[ObjectId, AllocatedObject[F, O]],
+    waiters: Map[WaiterId, Waiter[F, O]],
+  ) {
+
+    def findFreeObject: Option[(ObjectId, AllocatedObject[F, O])] =
+      objects.collectFirst {
```
t3hnar, line 49: I don't like performance characteristics of `findFreeObject`, it has a linear complexity. 
In often / worst case we will iterate through let's assume ~1k of entries to just find out that there is not a single one free…

t3hnar: why do you think so?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I'm not sure it's convenient to add `.timeout` on the caller side. If we want to support fiber cancellation we'll need to support it for both CE2 and CE3. I'm not familiar with how cancellation works in CE2 but I've glanced through the docs and it seems to be different from CE3. `ResourcePool` is private api and it's only used once, so I don't understand why we need this maintenance burden. My `acquireTimeout` implementation may be a crutch but it allows us to not think about CE2/CE3 compatibility.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,276 @@
...
+    log: Log[F],
+  )(
+    resource: Resource[F, O]
+  ): Resource[F, ResourcePool[F, O]] = {
+
+    type StateRef = Ref[F, State[F, O]]
+
+    def genUuid = Sync[F].delay(UUID.randomUUID())
+
+    def poolImpl(mutex: Mutex[F], stateRef: StateRef): ResourcePool[F, O] = {
```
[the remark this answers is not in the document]
edubrovski: `Mutex` is just a special case of `Semaphore(1)`, there's `Semaphore` in CE2.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
edubrovski: I think `scale = 0.5` is just a value we arrived to by trial and error :)

But in any case, you can tweak parameters in `ResourcePoolSpec` and see that the test completes very fast. I tried these params
```
val poolSize = 1000
val sleepDuration = 1.nanosecond
val timesToAcquire = 100_000
val idleTimeout = 1.second
val acquireTimeout = 30.seconds
```
and the test completed in a couple of seconds. So iterating through 1k elements in a map is not a bottleneck.

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
t3hnar: Do you want to say that passing `timeout` as argument is more "convenient" comparing to just using already provided api of `.timeout` ?

journal/src/main/scala/com/evolutiongaming/kafka/journal/util/ResourcePool.scala
```diff
@@ -0,0 +1,283 @@
...
+            case Some((id, obj)) =>
+              for {
+                _ <- waiter.complete((id, obj.value).asRight)
+                _ <- stateRef.update(_.markObjectAsUsed(id, obj))
+              } yield ()
+
+            case None =>
+              if (state.objects.size < poolSize) {
+                for {
+                  a <- allocateObject
```
t3hnar, line 125: one more reason to get rid of mutex, we block the whole thing while allocating/deallocating resource, and this would definitely be a problem

PR #501, opened 2023-07-26
Deduplication is fixed (workaround for https://issues.apache.org/jira/browse/CASSANDRA-18507)

DeviLab:


journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -268,15 +268,28 @@ object Journals {
...
-                val seqNr1 = a
-                  .seqNr
-                  .next[Option]
-                (seqNr1, Stream[F].single(a))
-              } else {
-                (seqNr.some, Stream[F].empty)
-              }
+            val empty = Option.empty[EventRecord[A]]
+
+            stream.map(Option(_)).concat(Stream.single(empty)).stateful(empty) {
```
dfakhritdinov, line 273: Existing implementation skips all events with `seqNr` less than `from` (line 271) while new one process full stream from the beginning, if I'm not missing anything

dfakhritdinov (changes requested):

dfakhritdinov:
Thanks for the fix, LGTM

dfakhritdinov (approved):

theide (approved):

PR #503, opened 2023-07-26
Deduplication is fixed (ce2)

DeviLab:


theide (approved):

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -268,15 +268,28 @@ object Journals {
...
+
+            stream.map(Option(_)).concat(Stream.single(empty)).stateful(empty) {
+              case (None, None) =>
+                None -> Stream.empty
+
+              case (None, Some(record)) =>
+                record.some.some -> Stream.empty
+
+              case (Some(prevRecord), Some(record)) if prevRecord.seqNr == record.seqNr =>
+                val logError = log.warn(s"Duplicated events found during recovery: $prevRecord and $record").toStream
```
t3hnar, line 281: it should be part of `debug` log, I don't like idea that because of database problems we generate such logs…

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -268,15 +268,28 @@ object Journals {
...
-                val seqNr1 = a
-                  .seqNr
-                  .next[Option]
-                (seqNr1, Stream[F].single(a))
-              } else {
-                (seqNr.some, Stream[F].empty)
-              }
+            val empty = Option.empty[EventRecord[A]]
+
+            stream.map(Option(_)).concat(Stream.single(empty)).stateful(empty) {
```
t3hnar, line 273: this all things should be part of `cassandra` specific layer and not of general kafka-journal

t3hnar, line 273: why did we remove previous check?

PR #505, opened 2023-08-03
Limit number of concurrent consumers

edubrovski:
This is an auxiliary PR, #500 is the main PR 

I tried to use `cats.effect.std.Backpressure` but adding timeout to it looked ugly. Implementation of Backpressure is pretty simple so I decided to write it myself using `bracket`. Then after some more experimentation I arrived at what I have now.


edubrovski:
We decided to go with the consumer pool so I'm closing this

PR #506, opened 2023-08-04
Reimplement deduplication logic:

dfakhritdinov:
on receiving records with duplicated `seqNr` from Kafka - take first one 
on receiving records with duplicated `seqNr` from Cassandra - raise error

### DIFF
--- a/journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
+++ b/journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
@@ -180,18 +180,6 @@ object Journals {
               eventual
                 .read(key, from)
                 .mapM { _.traverse(eventualRead.apply) }
-                .stateful(from) { case (seqNr, a) =>
-                  if (seqNr <= a.seqNr) {
-                    val seqNr1 = a
-                      .seqNr
-                      .next[Option]
-                    (seqNr1, Stream[F].single(a))
-                  } else {
-                    val msg = s"Data integrity violated: seqNr $seqNr duplicated in multiple records from eventual journal for key $key"
-                    val err = new JournalError(msg)
-                    (seqNr.some, err.raiseError[F, EventRecord[A]].toStream)
-                  }
-                }
             }
 
 
--- a/journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
+++ b/journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
@@ -180,6 +180,18 @@ object Journals {
               eventual
                 .read(key, from)
                 .mapM { _.traverse(eventualRead.apply) }
+                .stateful(from) { case (seqNr, a) =>
+                  if (seqNr <= a.seqNr) {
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/com/evolutiongaming/kafka/journal/JournalPerfSpec.scala
```diff
@@ -72,46 +72,48 @@ class JournalPerfSpec extends AsyncWordSpec with JournalSuite {
...
-      journalOf(eventualJournal).use(append)
+      for {
+        _ <- journal.pointer
+        _ <- expected.groupedNel(10).foldMap { events => journal.append(events).void }
+        _ <- appendNoise
+      } yield {}
     }
 
+    val appendToJournal = for {
+      _ <- awaitResources
```
t3hnar, line 111: could you please explain this a bit?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala
```diff
@@ -180,6 +180,18 @@ object Journals {
               eventual
                 .read(key, from)
                 .mapM { _.traverse(eventualRead.apply) }
+                .stateful(from) { case (seqNr, a) =>
```
t3hnar, line 183: I propose to move this into `EventualCassandra` implementation

t3hnar (changes requested):

tests/src/test/scala/com/evolutiongaming/kafka/journal/JournalPerfSpec.scala
dfakhritdinov: Sure

`def append(journals: Journals[IO]) = ???` function is one that pre-fill the journal before executing tests. In previous implementation it was executed before each test (ie twice) thus adding duplicated records to the journal. This PR changes eventual-journal behaviour by failing on data-inconvenience instead of skipping such records. To avoid failures that are not part of the test, I've changed pre-fill logic execution - its run once after `IntegrationSuite.start()` done to avoid having duplicates in the journal

t3hnar: maybe instead of hacking this around - we revert changes brought in hot-fix PR and just add new test case about this? otherwise this thing is getting really messy…

dfakhritdinov: I'm not hacking around here :) 

The test uses `EventualCassandra` by itself, even now after reverting temporal fix logic from `Journals` its still acts in the test. Previously the test was implemented in a way that `Journals` will deduplicate records which is not expected behaviour. To keep the test actual, I've changed pre-fill logic making it fear by executing only once.

If you see the change as over-complication I can propose adding docs (simple option) or reimplementing `JournalSuite` (probably) making pre-tests logic easily accessible from tests (it can take a while, I prefer to avoid it now)

t3hnar (approved):

PR #529, opened 2023-10-09
Execute one (instead # partitions) Cassandra query on head-cache cleanup

dfakhritdinov:


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/CassandraHelper.scala
```diff
@@ -92,4 +92,15 @@ object CassandraHelper {
   implicit val finiteDurationDecodeByName: DecodeByName[FiniteDuration] = {
     DecodeByName[DurationC].map(durationToFiniteDuration)
   }
+
+
+  implicit val setIntEncodeBbyName: EncodeByName[Set[Int]] = {
```
t3hnar, line 97: are we using this?

t3hnar (approved):

PR #532, opened 2023-11-01
Snapshot statements

rtar:


snapshot/src/main/scala/com/evolutiongaming/kafka/journal/BufferNr.scala
```diff
@@ -0,0 +1,37 @@
...
+sealed abstract case class BufferNr(value: Int) {
+  override def toString: String = value.toString
+}
+
+object BufferNr {
+
+  def listOf(size: Int): List[BufferNr] =
+    (0 until size).toList.map(fromIntUnsafe)
+
+  def fromIntUnsafe(value: Int): BufferNr =
```
t3hnar, line 19: 1. some validation? 
2. non negative?
3. where is safe version? :)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/CassandraHelper.scala
```diff
@@ -34,10 +34,17 @@ object CassandraHelper {
   }
 
 
+  implicit class RowOps(val self: Row) extends AnyVal {
+
+    def wasApplied: Boolean = self.getBool("[applied]")
```
t3hnar, line 39: `isApplied` ?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SchemaConfig.scala
```diff
@@ -12,6 +12,7 @@ final case class SchemaConfig(
   metaJournalTable: String = "metajournal",
   pointerTable: String = "pointer",
   pointer2Table: String = "pointer2",
+  snapshotTable: String = "snapshot",
```
t3hnar, line 15: just a note, akka currently uses `snapshots` table. We need to be 100% to not conflict with it :)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,182 @@
...
+import com.evolutiongaming.kafka.journal.eventual.EventualPayloadAndType
+import com.evolutiongaming.kafka.journal.eventual.cassandra.EventualCassandraConfig.ConsistencyConfig
+import com.evolutiongaming.scassandra.CassandraClusterOf
+
+import java.time.Instant
+
+object SnapshotCassandra {
+
+  // TODO: make it configurable
+  val BufferSize = 10
```
t3hnar, line 18: yes please!

snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
```diff
@@ -0,0 +1,19 @@
+package com.evolutiongaming.kafka.journal
+
+import java.time.Instant
+
+final case class SnapshotSelectionCriteria(
```
t3hnar, line 5: do you think we need this?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotStatements.scala
```diff
@@ -0,0 +1,328 @@
...
+       |buffer_nr INT,
+       |seq_nr BIGINT,
+       |timestamp TIMESTAMP,
+       |origin TEXT,
+       |version TEXT,
+       |metadata TEXT,
+       |payload_type TEXT,
+       |payload_txt TEXT,
+       |payload_bin BLOB,
+       |PRIMARY KEY ((id, topic, segment), buffer_nr))
```
t3hnar, line 33: buffer_idx ?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,182 @@
...
+      ): F[Unit] = {
+        val sortedSnapshots = savedSnapshots.toList.sortBy { case (_, (seqNr, timestamp)) => (seqNr, timestamp) }
+
+        val oldestSnapshot = sortedSnapshots.headOption
+        MonadThrow[F].fromOption(oldestSnapshot, SnapshotStoreError("Could not find an oldest snapshot")).flatMap {
+          oldestSnapshot =>
+            val (bufferNr, (deleteSnapshot, _)) = oldestSnapshot
+            val wasApplied = statements.updateRecord(key, segmentNr, bufferNr, insertSnapshot, deleteSnapshot)
+            wasApplied.flatMap { wasApplied =>
+              // TODO: consider adding circuit breaker here
```
t3hnar, line 100: could you please explain this comment?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,182 @@
...
+  private sealed abstract class Main
+
+  def apply[F[_]: MonadThrow](statements: Statements[F]): SnapshotStoreFlat[F] = {
+    new Main with SnapshotStoreFlat[F] {
+
+      // we do not use segments for now
+      val segmentNr = SegmentNr.min
+
+      def save(key: Key, snapshot: SnapshotRecord[EventualPayloadAndType]): F[Unit] = {
+        statements.selectMetadata(key, segmentNr).flatMap {
```
t3hnar, line 57: on the hot path given that buffer is full, do you need to query ALL snapshots?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotStatements.scala
```diff
@@ -0,0 +1,328 @@
...
+        } yield {
+
+          val seqNr = row.decode[SeqNr]
+          val bufferNr = row.decode[BufferNr]
+          val timestamp = row.decode[Instant]("timestamp")
+
+          (bufferNr, (seqNr, timestamp))
+        }
+
+        rows.toList.map(_.toMap)
```
t3hnar, line 218: minor, not sure `toMap` should be part of `statement` :)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,182 @@
...
+        }
+      }
+
+      def replace(
+        key: Key,
+        segmentNr: SegmentNr,
+        savedSnapshots: Map[BufferNr, (SeqNr, Instant)],
+        insertSnapshot: SnapshotRecord[EventualPayloadAndType]
+      ): F[Unit] = {
+        val sortedSnapshots = savedSnapshots.toList.sortBy { case (_, (seqNr, timestamp)) => (seqNr, timestamp) }
```
t3hnar, line 92: no need to sort collection, just `minBy`

t3hnar: is it correct to include `timestamp` here?

t3hnar (changes requested):

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
rtar: Technically, it should never happen, i.e. there should not be two snapshots with the same `seqNr`. :thinking:

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/CassandraHelper.scala
rtar: I named it `wasApplied`, because it is called like that in Cassandra driver, so I thought it might ease discoverability a bit. But `isApplied` sounds fine too.

https://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/ResultSet.html#wasApplied--

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SchemaConfig.scala
rtar: I did not think about it, to be honest. Other name is required here, indeed.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
rtar: I need this information to find the oldest snapshot to be replaced. There are might be the ways to optimize it, though. :thinking:

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotStatements.scala
rtar: Fixed

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SchemaConfig.scala
rtar: How about "snapshot_buffer"?

snapshot/src/main/scala/com/evolutiongaming/kafka/journal/BufferNr.scala
rtar: Fixed

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/SnapshotCassandra.scala
rtar: Fixed

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+      }
+
+      def update(
+        key: Key,
+        savedSnapshots: Map[BufferNr, (SeqNr, Instant)],
+        insertSnapshot: SnapshotRecord[EventualPayloadAndType]
+      ): F[Unit] = {
+        val sortedSnapshots = savedSnapshots.toList.sortBy { case (_, (seqNr, timestamp)) => (seqNr, timestamp) }
```
dfakhritdinov, line 102: What's the point in sorting snapshots if later sorted list will be used only once:

```
sortedSnapshots.find ...
```
?

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,172 @@
...
+          }
+        }
+      }
+
+      def replace(
+        key: Key,
+        savedSnapshots: Map[BufferNr, (SeqNr, Instant)],
+        insertSnapshot: SnapshotRecord[EventualPayloadAndType]
+      ): F[Unit] = {
+        val oldestSnapshot = savedSnapshots.toList.minByOption { case (_, (seqNr, timestamp)) => (seqNr, timestamp) }
```
dfakhritdinov, line 84: Are you expecting to have few snapshots with same `seqNr`? I do not understand use-case for it, could you please share more details

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
```diff
@@ -0,0 +1,172 @@
...
+            wasApplied.flatMap { wasApplied =>
+              // TODO: consider adding circuit breaker here
+              if (wasApplied) ().pure[F] else save(key, insertSnapshot)
+            }
+        }
+      }
+
+      def load(key: Key, criteria: SnapshotSelectionCriteria): F[Option[SnapshotRecord[EventualPayloadAndType]]] =
+        for {
+          savedSnapshots <- statements.selectMetadata(key)
```
dfakhritdinov, line 119: Would it be possible to select snapshot without querying metadata first? I did not check all details, but criteria looks like something that can be expressed in terms of CQL

rtar: I was trying to avoid `ALLOW FILTERING` here.

Saying that, as we do have primary columns in `WHERE` clause then CQL might, indeed, work here. :thinking:

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
rtar: It should never have several snapshots with the same `seqNr`, but it seemed like a good idea to be a little bit defensive here.

I.e., if there is a bug or something, it will still filter by `timestamp` and delete an oldest one, not a random one.

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
rtar: The point is to update the oldest one if several snapshots are found.

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotCassandra.scala
dfakhritdinov: Would it be possible to log (or report any other way) the case of duplicated `seqNr`? Please resolve the conversation if its not easily possible

snapshot-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/snapshot/cassandra/SnapshotStatements.scala
```diff
@@ -0,0 +1,309 @@
...
+           |seq_nr,
+           |timestamp,
+           |origin,
+           |version,
+           |payload_type,
+           |payload_txt,
+           |payload_bin,
+           |metadata)
+           |VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+           |IF NOT EXISTS
```
dfakhritdinov, line 66: As was pointed out by @migesok, [mixing LWT and normal operations can result in errors](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlLtwtTransactions.html):

> Lightweight transactions will block other lightweight transactions from occurring, but will not stop normal read and write operations from occurring. Lightweight transactions use a timestamping mechanism different than for normal operations and mixing LWTs and normal operations can result in errors. If lightweight transactions are used to write to a row within a partition, only lightweight transactions for both read and write operations should be used. This caution applies to all operations, whether individual or batched.

@rtar, could we instead of LWT rely on Single Writer Principle provided by Akka?

rtar: This paragraph only applies to the write operations and the read operations requiring a strong consistency requirement.

It should be safe to read the data, given you are okay with eventual consistency or if you use SERIAL consistency level while reading it.

The documentation is very confusing in that area.

Saying that, after considering this for a longer time, I decided that LWTs are totally unnecessary here, indeed, given a single writer is guaranteed externally, as you suggested. 

I already started a work to remove LWTs. I thought I shared this with @migesok, but I think I confused him by saying that writing snapshot will not be guaranteed. It will be guaranteed, actually, given the single writer is ensured.

dfakhritdinov: Thanks for explanation 🙇 , I was not aware of your plans to move on with SWP instead of LWT thus will be happy to review final version of the snapshot store

rtar: LWTs are disabled by default, now.

PR #541, opened 2023-12-15
Extract consistency settings into a separate class

rtar:
The idea is to refer to it from different configuration case classes, i.e. from `EventualCassandraConfig`, `SnapshotCassandraConfig` and `CassandraHealthCheck`.

### DIFF
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraConfig.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraConfig.scala
@@ -49,8 +49,7 @@ object EventualCassandraConfig {
   implicit val configReaderEventualCassandraConfig: ConfigReader[EventualCassandraConfig] = deriveReader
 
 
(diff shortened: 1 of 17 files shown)


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraConfig.scala
```diff
@@ -40,41 +40,12 @@ final case class EventualCassandraConfig(
...
+  consistencyConfig: ConsistencyConfig = ConsistencyConfig.default)
 
 object EventualCassandraConfig {
 
   val default: EventualCassandraConfig = EventualCassandraConfig()
 
   implicit val configReaderEventualCassandraConfig: ConfigReader[EventualCassandraConfig] = deriveReader
 
-
-  final case class ConsistencyConfig(
```
t3hnar, line 52: please at leave empty dummy class with deprecation and a link - where to look for the replacement

rtar: I fixed it, but I had to rename the class to `CassandraConsistencyConfig` to avoid a naming clash.

Should I do the same in https://github.com/evolution-gaming/kafka-journal/pull/539? :thinking:

t3hnar (approved):

PR #555, opened 2024-01-23
Fix the flakiness of DamperTest

rtar:
I wrote a lot of documentation to understand how it works, and debugged it a lot.

It looks like the problem is in the test, which uses `Ref` to specify delays to be used (instead of relying on `acquired` argument).

If the delay between starting two threads is too small then concurrency race happens and `delayOf` function is called twice, when starting `fiber0` and `fiber1`.

I did not convince myself that it is a wrong behavior, so I just increased a delay in the test, so concurrent race does not happen.


dfakhritdinov (approved):
Great scaladocs, thanks

PR #562, opened 2024-02-02
Add ring buffer index implementation

rtar:
There will be several rows in Cassandra per a single snapshot primary key. Each of these rows will differ by an index, which this class is going to represent.

### DIFF
--- a/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/BufferNr.scala
+++ b/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/BufferNr.scala
@@ -1,0 +1,65 @@
+package com.evolutiongaming.kafka.journal
+
+import cats.syntax.all._
+import cats.{Applicative, Id}
+import com.evolutiongaming.kafka.journal.util.Fail
+import com.evolutiongaming.kafka.journal.util.Fail.implicits._
+import com.evolutiongaming.scassandra._
+
+/** Snapshot index in a stored ring buffer */
+sealed abstract case class BufferNr(value: Int) {
+  override def toString: String = value.toString
+}
+
+object BufferNr {
+
+  val min: BufferNr = BufferNr.fromIntUnsafe(0)
+  val max: BufferNr = BufferNr.fromIntUnsafe(Int.MaxValue)
+
+  /** Create all list of buffer indicies of a given size.
+    *
+    * I.e. for `size = 3` the following list will be created:
+    * {{{
+    * List(BufferNr(0), BufferNr(1), BufferNr(2))
+    * }}}
+    */
+  def listOf(size: Int): List[BufferNr] =
+    (0 until size).toList.map(fromIntUnsafe)
+
+  private def fromIntUnsafe(value: Int): BufferNr =
+    new BufferNr(value) {}
+
+  /** Create `BufferNr` out of a value or fail it is out of an allowed range.
+    *
+    * A returned value may be reused to minimize number of allocations.
+    */
+  def of[F[_]: Applicative: Fail](value: Int): F[BufferNr] = {
+    if (value < min.value) {
+      s"invalid BufferNr of $value, it must be greater or equal to $min".fail[F, BufferNr]
+    } else if (value > max.value) {
+      s"invalid BufferNr of $value, it must be less or equal to $max".fail[F, BufferNr]
+    } else if (value === min.value) {
+      min.pure[F]
+    } else if (value === max.value) {
+      max.pure[F]
+    } else {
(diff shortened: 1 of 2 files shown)


Z1kkurat:
I'm lacking some knowledge here. Could you please elaborate on `There will be several rows in Cassandra per a single snapshot primary key.` - why is it so?

rtar:
@Z1kkurat 

The current Akka Persistence Cassandra plugin most people use has the following primary key:
```
PRIMARY KEY (persistence_id, sequence_nr))
```
The new plugin will have the following key instead:
```
PRIMARY KEY ((id, topic), buffer_idx))
```
The problem is that the existing plugin creates a tombstone on every new snapshot update (because it has to create a new row with a new `sequence_nr`. The consequence is not only the data/performance waste, but also a potential incident, when the snapshot cannot be restored at all, because tombstone limit was reached for a query.

The new plugin is intended to solve the issue by overwriting the older snapshot in the exiting row.

This, however, could be dangerous in case of concurrency issues. I.e. the newer snapshot version may be overwritten by an older one. Also, having some older snapshots might be beneficial in case of the data corruption incidents.

That why the big idea is to have a small ring buffer of snapshots and overwrite the oldest once the ring buffer is full.

Z1kkurat:
> @Z1kkurat
> 
> The current Akka Persistence Cassandra plugin most people use has the following primary key:
> 
> ```sql
> PRIMARY KEY (persistence_id, sequence_nr))
> ```
> 
> The new plugin will have the following key instead:
> 
> ```sql
> PRIMARY KEY ((id, topic), buffer_idx))
> ```
> 
> The problem is that the existing plugin creates a tombstone on every new snapshot update (because it has to create a new row with a new `sequence_nr`). The consequence is not only the data/performance waste, but also a potential incident, when the snapshot cannot be restored at all, because tombstone limit was reached for a query.
> 
> The new plugin is intended to solve the issue by overwriting the older snapshot in the exiting row.
> 
> This, however, could be dangerous in case of concurrency issues. I.e. the newer snapshot version may be overwritten by an older one. Also, having some older snapshots might be beneficial in case of the data corruption incidents.
> 
> That why the big idea is to have a small ring buffer of snapshots and overwrite the oldest once the ring buffer is full.

Thank you for a thorough explanation, I appreciate it! A couple more questions if you don't mind:  
1. If I understood you correctly, we'd always have X rows for each `(id, topic)` in Cassandra where X is the length of the ring buffer. Did we evaluate how this would impact the volume of data kept? Currently, if older tombstones are deleted fast enough during compaction, we'd have a small enough footprint (I doubt it happens in real life, but anyway). With the ring buffer, we'd constantly have a fixed number of rows
2. Can there be a race for writing at the same next spot in the buffer? Will we be able to mitigate that?

These questions don't affect the PR though, so I'll approve it anyway

Z1kkurat (approved):

rtar:
@Z1kkurat

> If I understood you correctly, we'd always have X rows for each (id, topic) in Cassandra where X is the length of the ring buffer. Did we evaluate how this would impact the volume of data kept? Currently, if older tombstones are deleted fast enough during compaction, we'd have a small enough footprint (I doubt it happens in real life, but anyway). With the ring buffer, we'd constantly have a fixed number of rows

The deletion will still be available in a new plugin. Deleting a snapshot will create a tombstone too. But the number of tombstones per `(id, topic)` will be limited to that small X we choose. I.e. if we set X = 10 and do insert/delete 1000 of snapshots, we will still have 1 alive snapshot and 9 tobmstones before compaction process kicks in, rather than 999 snapshot like with Akka Pesistence Cassandra plugin.

> Can there be a race for writing at the same next spot in the buffer? Will we be able to mitigate that?
Yes, there can be such a race. 

The current prototype uses lightweight transactions (LWTs) to detect the situation and retry the insert into a different place. I am bit worried about the use of LWTs, but I could not figure out the simple workaround yet, except for reimplementing LWTs on a user level.

The current implementation idea could be found here:
https://github.com/evolution-gaming/kafka-journal/pull/532/files#diff-aad9a40d5cc6a530195dfbca0646615201670dfb39bfe3469658b967f20832cbR72

Here you can find some of the corner cases tested:
https://github.com/evolution-gaming/kafka-journal/pull/532/files#diff-38c946f0eede7c202c15bb08d5b42017f07c1d303846171a510a61274d17acf7R33

(I am deconstructing this large PR into smaller PRs, so you will get to review these classes eventually too)

One alternative idea to explore could be to include some loaned writerId into a primary key. This could ensure the snapshots are never overwritten while still limiting number of tombstones, but avoiding the use of LWT mechanism.

PR #563, opened 2024-02-05
Add a class to pass SeqNr and snapshot payload in

rtar:
The main question here is if we want to have `payload` as an `Option` or require it always to be filled.

This idea of `Option` was taken from a similar `Event` case class, but I am not sure it applies the same way to a snapshot.

### DIFF
--- a/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/Snapshot.scala
+++ b/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/Snapshot.scala
@@ -5,9 +5,9 @@ package com.evolutiongaming.kafka.journal
   * @param seqNr
   *   Snapshot sequence number.
   * @param payload
-  *   Actual contents of a snapshot. The value is never `None` under normal circumstances.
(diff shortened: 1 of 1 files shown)


Z1kkurat:
I struggle to imagine a case where we'd need a snapshot without a state inside. It looks like a tombstone, but we don't need those for persistent storage such as Cassandra, do we? We either have a snapshot with a state, or we delete the row altogether

rtar:
@Z1kkurat  I copied from `Event` structure.

I now suspect it was added there as part of the https://github.com/evolution-gaming/kafka-journal/commit/c6d0543b908b1475e8a6bdf4eebc45d5d8733dd5 commit to ensure backwards compatibility with the previously stored binary snapshot.

I will rework it to remove `Option` now.

CC: @dfakhritdinov

rtar:
I have removed `Option` now (and also removed it from a large PR, which I am deconstructing into these smaller PRs, but you do not have to look there: https://github.com/evolution-gaming/kafka-journal/pull/532).

Z1kkurat (approved):

PR #566, opened 2024-02-09
Implement SnapshotRecord class

rtar:
The idea is to have a class that will be directly mapped to Cassandra columns in a table. The class is inspired by a similar `EventRecord` class. Roughly speaking it is snapshot seqNr/payload + snapshot metadata.

Once again, I am not sure if `version` and `origin` fields should be optional like they are in `EventRecord`.

It may make some sense from uniformity point of view, and for potential ability to save space if there are a lot of snapshots. But I am not sure.


dfakhritdinov:
I think its fine to have optional `origin` and `version` as those fields not used in journal logic

dfakhritdinov (approved):

PR #570, opened 2024-02-14
Add SnapshotStoreFlat interface

rtar:
This is intended to be a main entry point to a new snapshotter.
    
The main considerations were to keep it recognizable and easy to use from a Akka Persistence Plugin, while having no dependencies on Akka itself.
 
I also tried to make it look similar to `ReplicatedJournalFlat` and `EventualJournal` for sake of uniformity of the codebase.

See also https://github.com/evolution-gaming/kafka-journal/pull/532/files#r1412205449 for a comment from @t3hnar

### DIFF
--- a/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
+++ b/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
@@ -2,6 +2,13 @@ package com.evolutiongaming.kafka.journal
 
 import java.time.Instant
 
+/** Type safer version of `akka.persistence.SnapshotSelectionCriteria`.
+  * 
+  * The important consideration when using this class is to always use named
+  * parameters to consruct the instances of the class. Otherwise it is too
+  * easy to confuse `maxSeqNr` and `minSeqNr`, especially, given the fact,
+  * they are given in a reverse order in a constructor.
+  */
 final case class SnapshotSelectionCriteria(
   maxSeqNr: SeqNr = SeqNr.max,
   maxTimestamp: Instant = Instant.MAX,
--- a/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
+++ b/snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
@@ -1,0 +1,19 @@
+package com.evolutiongaming.kafka.journal
+
+import java.time.Instant
+
+final case class SnapshotSelectionCriteria(
+  maxSeqNr: SeqNr = SeqNr.max,
+  maxTimestamp: Instant = Instant.MAX,
+  minSeqNr: SeqNr = SeqNr.min,
+  minTimestamp: Instant = Instant.MIN
+)
+
+object SnapshotSelectionCriteria {
+
+  val All: SnapshotSelectionCriteria = SnapshotSelectionCriteria()
+
+  def one(seqNr: SeqNr): SnapshotSelectionCriteria =
+    SnapshotSelectionCriteria(maxSeqNr = seqNr, minSeqNr = seqNr)
+
+}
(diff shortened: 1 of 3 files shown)


snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotSelectionCriteria.scala
```diff
@@ -0,0 +1,26 @@
...
+import java.time.Instant
+
+/** Type safer version of `akka.persistence.SnapshotSelectionCriteria`.
+  * 
+  * The important consideration when using this class is to always use named
+  * parameters to consruct the instances of the class. Otherwise it is too
+  * easy to confuse `maxSeqNr` and `minSeqNr`, especially, given the fact,
+  * they are given in a reverse order in a constructor.
+  */
+final case class SnapshotSelectionCriteria(
```
dfakhritdinov, line 12: Just to be on a same page: new akka-effect persistence API does not accept the criteria https://github.com/evolution-gaming/akka-effect/blob/262cc7b70a2f7419644d9c4b7644b4b1888f5733/persistence-api/src/main/scala/com/evolutiongaming/akkaeffect/persistence/SnapshotStore.scala#L19

thus in integration always will be used `SnapshotSelectionCriteria.All`, IMO its okay

snapshot/src/main/scala/com/evolutiongaming/kafka/journal/SnapshotStoreFlat.scala
```diff
@@ -0,0 +1,58 @@
+package com.evolutiongaming.kafka.journal
+
+import com.evolutiongaming.kafka.journal.eventual.EventualPayloadAndType
+
+/** Snapshot storage for Kafka Journal.
+  *
+  * Uses `Key` instead of `persistenceId`, similar to [[ReplicatedJournalFlat]]
+  * and [[EventualJournal]].
+  */
+trait SnapshotStoreFlat[F[_]] {
```
dfakhritdinov, line 10: Am I right that later you're planning to introduce non-flat trait?

I'm asking cos earlier you've added scaladocs where explicitly mentioned:
> The flat interface was replaced by hierarchical [[ReplicatedJournal]] for
> all means and purposes except for the unit tests, and should not be used
> directly anymore.

Do we need flat trait at all?

I've seen reasoning in PR description:
> I also tried to make it look similar to ReplicatedJournalFlat and EventualJournal for sake of uniformity of the codebase.

and would vote for keeping the implementation simple. Might be better would be dropping other flat traits then introducing new one? BUT **I don't insist**, feel free to resolve the comment 👍

rtar: I was thinking about having the flat one only.

The hierarchical structure of `ReplicatedJournal` kind of makes sense, because it is writing to Kafka and therefore has to deal with topics and partitions.

`SnapshotStoreFlat` implementation uses Cassandra only and just uses `(topic, id)` tuple as a key, so hierarchy won't add much except a bit more convoluted interface. That why, I suspect, @t3hnar never made `EventualJournal` hierarchical, because it also deals with Cassandra only.

I could rename it to `SnapshotStore` therefore, though I am a bit weary that there is `akka.persistence.snapshot.SnapshotStore` already and if this is going to be too confusing. May be some other name could be more appropriate.

t3hnar: minor: I saw others are naming their impls as `SnapshotStore`, package name should make it distinguishable :)

dfakhritdinov: IMO: I struggled yesterday with 3 classes from different libs (and packages) all named as `EventSourcedOf` :)

dfakhritdinov: @rtar, I'm fine with `SnapshotStore`, hope the only place where one would case of distinguish them is akka-effect/persistence module

PR #587, opened 2024-02-29
Add `sbt-version-policy` to check for backwards compatibility issues

rtar:
### Idea

This PR integrates  `sbt-version-policy` into a build process.

One can now run `sbt versionPolicyCheck` to validate if the upcoming version (3.3.9 at the the time of writing) is binary compatible with the previous one.

That task will fail, because the version is not backwards compatible at the moment of writing. That why the task is not integrated to CI yet.

The version is made compatible in https://github.com/evolution-gaming/kafka-journal/pull/586, which will be split to smaller PRs for an easier review (this is the first such a smaller PR).

### Details

I had to update `build.sbt` with a temporary hack which creates a separate `mima-classes` directory for journal and `eventual-cassandra` modules to make sure that `core` and `cassandra` module classes are included as in 3.3.8 version these classes were located together. That hack won't be necessary in 3.3.10 anymore.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,8 @@ lazy val commonSettings = Seq(
     "org.scala-lang.modules" %% "scala-java8-compat" % "always",
     "org.scala-lang.modules" %% "scala-xml"          % "always"),
   autoAPIMappings := true,
-  versionScheme := Some("early-semver"))
+  versionScheme := Some("early-semver"),
+  versionPolicyIntention := Compatibility.BinaryCompatible)
 
 
 lazy val root = (project in file(".")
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -92,6 +108,7 @@ lazy val journal = (project in file("journal")
     `cassandra-driver`,
     scassandra,
     scache,
+    `cassandra-sync`,
```
Z1kkurat, line 111: is it used?

rtar: No, it is not used. I was the one removing it, initially, in previous commits.

I have added it back, because lack of it breaks the backwards compatibility.

I.e. the library depending on `kafka-journal` will not get it anymore.

MiMa detects it and complains.

Z1kkurat (approved):

PR #610, opened 2024-04-24
update build infrastructure

mr-git:


journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/EventualJournal.scala
```diff
@@ -356,7 +356,7 @@ object EventualJournal {
         def offset(topic: Topic, partition: Partition): F[Option[Offset]] = {
           self
             .offset(topic, partition)
-            .handleErrorWith { a =>
+            .handleErrorWith { (a) =>
```
dfakhritdinov, line 359: is it scala3 requirement? Can we change scalafmt config somehow to avoid braces for single untyped argument? I strongly disagree with such style

journal/src/test/scala/com/evolutiongaming/kafka/journal/JournalSpec.scala
```diff
@@ -675,7 +675,7 @@ object JournalSpec {
             val partitionOffset = PartitionOffset(partition, record.offset)
             EventRecord(action, event.map(eventualWrite(_).get), partitionOffset, events1.metadata)
           }
-          copy(events = events.enqueue(batch.toList), offset = offset.some)
+          copy(events = events.enqueueAll(batch.toList), offset = offset.some)
```
dfakhritdinov, line 678: @mr-git could you please elaborate this change

journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/EventualJournal.scala
mr-git: https://docs.scala-lang.org/scala3/guides/migration/incompat-syntactic.html#parentheses-around-lambda-parameter - that is the syntax change

journal/src/test/scala/com/evolutiongaming/kafka/journal/JournalSpec.scala
mr-git: `enqueue` is deprecated with suggestion to use `enqueueAll`:
https://github.com/scala/scala/blob/a5270194b6eec9678a79e2f88a83d83aa6b92de1/src/library/scala/collection/immutable/Queue.scala#L152

journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/EventualJournal.scala
dfakhritdinov: The doc says

> When followed by its type, the parameter of a lambda is now required to be enclosed in parentheses. The following piece of code is invalid.

I see no type here thus expect to have to braces

mr-git: indeed - reverted

dfakhritdinov (approved):

PR #630, opened 2024-06-26
SHRIKE-52: make data integrity validation configurable

dfakhritdinov:


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
```diff
@@ -99,7 +156,15 @@ object EventualCassandra {
     * The implementation itself is abstracted from the calls to Cassandra which
     * should be passed as part of [[Statements]] parameter.
     */
-  def apply[F[_]: MonadThrow](statements: Statements[F]): EventualJournal[F] = {
+  @deprecated("Use of1 instead", "3.4.0")
```
mr-git, line 159: maybe?
```suggestion
  @deprecated("Use apply1 instead", "3.4.0")
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
```diff
@@ -120,28 +185,35 @@ object EventualCassandra {
...
-                } else {
-                  val msg = s"Data integrity violated: seqNr $seqNr duplicated in multiple records from eventual journal, key $key"
-                  val err = new JournalError(msg)
-                  (seqNr.some, err.raiseError[F, EventRecord[EventualPayloadAndType]].toStream)
+                .map { case (record, _) => record }
+
+            if (dataIntegrity.seqNrUniqueness) {
+              events
+                .stateful(from) { case (seqNr, record) =>
+                  if (seqNr <= record.seqNr) {
```
mr-git, line 203: why include check for equality? shouldn't it be just `<` instead of `<=`?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
```diff
@@ -343,4 +343,17 @@ object Journal {
...
+  }
+
+  object DataIntegrityConfig {
+
+    private case class Implementation(seqNrUniqueness: Boolean) extends DataIntegrityConfig 
+
+    val Default: DataIntegrityConfig = Implementation(seqNrUniqueness = true)
+
+    implicit val configReaderDataIntegrityConfig: ConfigReader[DataIntegrityConfig] = deriveReader[Implementation].map(a => a: DataIntegrityConfig)
+  }
```
mr-git, line 358: Is `seqNrUniqueness: Boolean` the best we can do? Nicely named ADTs might be more expressive: `CheckSeqNrGrowing` (default) and `IgnoreChecks` (new) 

Frankly, we never check for `Uniqueness` - we check that new `seqNr` larger than previously observed one, but we never cehck for delta to be exactly one, do we? We could add `StrictCheckSeqNr` to do checks for exactly `+1` though...

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournalConfig.scala
```diff
@@ -44,6 +44,10 @@ import scala.concurrent.duration._
   *   vice-versa. This parameter is only relevant if default [[KafkaJournal]] is
   *   used, i.e. it is not taken into account if Circe JSON or other custom
   *   serialization is used.
+  * @param consumerPool
+  *   Kafka consumer pool configuration.
+  * @dataIntegrity
```
mr-git, line 49:
```suggestion
  * @param dataIntegrity
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
dfakhritdinov: Would be better asking author of the code, but AFAIK the reason is in `seqNr` value - its _not_ last seen sequence number but rather next one expected:
```
                  if (seqNr <= record.seqNr) {
                    val seqNr1 = record
                      .seqNr
                      .next[Option]
                    (seqNr1, Stream[F].single(record))
                  }
```
In case of no gaps in sequence numbers would be enough having `seqNr == record.seqNr`, but in reality gaps are happening regularly thus KJ will accept them and delegate accepting/rejecting event to the application.

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
dfakhritdinov: I'm fine with naming and strongly against ADT for true|false values, please do not increase complexity without reason. 

> Frankly, we never check for Uniqueness - we check that new seqNr larger than previously observed one, but we never cehck for delta to be exactly one, do we? We could add StrictCheckSeqNr to do checks for exactly +1 though...

KJ should not check that current `seqNr` is exactly +1 of previous due to possible gaps in the sequence. Now final decision of accepting or rejecting event delegated to the application instead of been forced in KJ.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
mr-git: why gaps happen?

persistence/src/main/scala/akka/persistence/kafka/journal/KafkaJournalConfig.scala
```diff
@@ -56,6 +60,7 @@ final case class KafkaJournalConfig(
   callTimeThresholds: CallTimeThresholds = CallTimeThresholds.default,
   jsonCodec: KafkaJournalConfig.JsonCodec = KafkaJournalConfig.JsonCodec.Default,
   consumerPool: ConsumerPoolConfig = ConsumerPoolConfig.Default,
+  dataIntegrity: DataIntegrityConfig = DataIntegrityConfig.Default,
```
mr-git, line 63: seems that MiMa check wants us to add custom constructor `apply` without `dataIntegrity` - probably we can do that in companion object...

dfakhritdinov: no, I disabled MiMa for now. Will bring it back after merging the PR

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
dfakhritdinov: Again, the question is not related to the PR. Lets not block it by side discussions.

I suggest asking @theide as one having much more experience in using KJ :D

mr-git (approved):

PR #639, opened 2024-07-09
Bump play-json-jsoniter to 1.1.1

odbc:
Play-json-jsoniter [diff](https://github.com/evolution-gaming/play-json-tools/compare/v0.10.2...v1.1.1)

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-ThisBuild / version := "3.4.1-SNAPSHOT"
+ThisBuild / version := "4.0.0"
--- a/build.sbt
+++ b/build.sbt
@@ -71,8 +71,6 @@ lazy val `scalatest-io` = (project in file("scalatest-io")
 lazy val core = (project in file("core")
   settings (name := "kafka-journal-core")
   settings commonSettings
-  // The following line should be removed once 3.3.9 is released.
-  settings (versionPolicyCheck / skip := true)
   dependsOn (`scalatest-io` % Test)
   settings (libraryDependencies ++= Seq(
     Akka.actor,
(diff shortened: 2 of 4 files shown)


version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "3.4.1-SNAPSHOT"
+ThisBuild / version := "4.0.0"
```
mr-git, line 1: @dfakhritdinov, I suppose we are not ready for this, yet :)

@odbc, we have a lot more of cleanup and changes planned for 4.0.0 and we'd like to postpone release of 4.0.0 for a while. How `1.1.1` differs from `0.10.2` besides package and version? is it otherwise binary compatible?

plokhotnyuk: @odbc @mr-git @dfakhritdinov What do you think about contribution of play-json-jsoniter to jsoniter-scala repo? 

It can fix missing builds for Scala 3 and Scala Native.

Also, per commit MiMa checks will guarantee backward binary compatibility for minor versions and both directions (backward and forward) for path versions.

/cc @t3hnar

odbc: As it should be actually compatible. Looking at [diff](https://github.com/evolution-gaming/play-json-tools/compare/v0.10.2...v1.1.1) I noticed only some changes in signatures of internal API ([JsonValueCodecJsValue.apply](https://github.com/evolution-gaming/play-json-tools/compare/v0.10.2...v1.1.1#diff-4817a3b1e9fcb97d56625798ba3aee4c91d4b9747165c9a9e4c9152187045eb9L10))

But the versions compatibility checker considers `play-json-jsoniter` versions binary incompatible and makes the PR red

dfakhritdinov: Its not only about making the PR red 😞 

Almost all Evolution libs/apps are using semver thus the bump of `play-json-jsoniter` version will end-up with bumping it everywhere where its present (transitively). I agree with @mr-git and prefer to combine major version bump with some other planned maintenance.

@odbc, how urgent is the PR?

odbc: > @odbc, how urgent is the PR?

It can wait for your major update

pizzaeueu: Hi there!

@plokhotnyuk 
I can try to help with `play-json-jsoniter` -> `jsoniter-scala` contribution if nobody minds

cc: @dfakhritdinov

dfakhritdinov:
closed in favor of https://github.com/evolution-gaming/kafka-journal/pull/664

PR #641, opened 2024-07-10
expose java Kafka producer metrics

dfakhritdinov:


persistence/src/main/scala/akka/persistence/kafka/journal/JournalAdapter.scala
```diff
@@ -68,7 +68,8 @@ object JournalAdapter {
         val clientId = clientIdOf(config.journal.kafka.producer.common)
         metrics(clientId)
       }
-      KafkaProducerOf[F](producerMetrics)
+      val prefix = clientIdOf(config.journal.kafka.consumer.common)
```
migesok, line 71: client ID is already included in the generated labels

persistence/src/main/scala/akka/persistence/kafka/journal/JournalAdapter.scala
```diff
@@ -68,7 +68,8 @@ object JournalAdapter {
...
+      KafkaProducerOf.apply1[F](prefix, producerMetrics)
```
migesok, line 72: there should be a way to disable producer metrics without disabling overall journal metrics

dfakhritdinov: why should the metrics be so configurable? I see not much sense in it

persistence/src/main/scala/akka/persistence/kafka/journal/JournalAdapter.scala
dfakhritdinov: in what kind of labels the ID already included?

dfakhritdinov:
The PR must be re-implemented after https://github.com/evolution-gaming/skafka/pull/430 will be merged. New `skafka` feature will allow much smoother integration

persistence/src/main/scala/akka/persistence/kafka/journal/JournalAdapter.scala
migesok: Please try to run it locally - client_id is already included in the labels generated by the collector.
I suspect if you generate the prefix this way, it won't even mention that those are "kafka producer" metrics.

persistence/src/main/scala/akka/persistence/kafka/journal/JournalAdapter.scala
migesok: If they name clash with other things, it's good to be able to disable them.

PR #642, opened 2024-07-10
Expose new temporary metrics when `pointer` table is used as fallback 

### LINKED ISSUE #636: expose new temporary metric(s) on when pointer table is used as fallback
Part of epic: #635 

Have to expose temporary metric to understand, if `pointer` table provides useful data as fallback. 
Expectation is that it does not and that it is safe to remove its usage.

mr-git:
Fixes #636

### DIFF
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
@@ -4,10 +4,12 @@ import cats.data.NonEmptyList as Nel
 import cats.effect.syntax.all.*
 import cats.effect.{Async, Ref, Resource, Sync}
 import cats.syntax.all.*
-import cats.{Monad, Parallel}
+import cats.{Applicative, Monad, Parallel}
 import com.evolutiongaming.catshelper.ParallelHelper.*
-import com.evolutiongaming.catshelper.{LogOf, MeasureDuration, ToTry}
+import com.evolutiongaming.catshelper.{Log, LogOf, MeasureDuration, ToTry}
 import com.evolutiongaming.kafka.journal.eventual.*
+import com.evolutiongaming.kafka.journal.eventual.ReplicatedKeyJournal.Changed
+import com.evolutiongaming.kafka.journal.eventual.cassandra.ReplicatedCassandra.MetaJournalStatements.ByKey
 import com.evolutiongaming.kafka.journal.eventual.cassandra.SegmentNr.implicits.*
 import com.evolutiongaming.kafka.journal.util.CatsHelper.*
 import com.evolutiongaming.kafka.journal.util.Fail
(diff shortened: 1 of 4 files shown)


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -52,60 +56,104 @@ object ReplicatedCassandra {
...
+
+      def topics: F[SortedSet[Topic]] = {
         statements
           .selectTopics2()
           .parProduct(statements.selectTopics())
-          .map { case (a, b) => a ++ b }
+          .flatMap {
+            case (a, b) =>
+              val meterAndLog = Applicative[F].whenA(b.diff(a).nonEmpty) {
+                val diff = b.diff(a)
```
dfakhritdinov, line 82: Minor thing: you're doing `b.diff(a)` twice, might be reasonable to move
```
val diff = b.diff(a)
```
out of `.whenA(...)` scope and use in in the condition and body.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -52,60 +56,104 @@ object ReplicatedCassandra {
...
+      val unit: F[Unit] = Applicative[F].unit
+
+      def topics: F[SortedSet[Topic]] = {
         statements
           .selectTopics2()
           .parProduct(statements.selectTopics())
-          .map { case (a, b) => a ++ b }
+          .flatMap {
+            case (a, b) =>
+              val meterAndLog = Applicative[F].whenA(b.diff(a).nonEmpty) {
```
dfakhritdinov, line 81: Minor thing:
Am I right that executing `meterAndLog` with `metrics == None` and `log == None` pointless?

Might be you can do something like:
```
metrics.product(log).traverse { case (metrics, log) =>
  ???
}
```

dfakhritdinov (approved):

dfakhritdinov:
@mr-git please do not use names with symbols that require quoting in bash. 

Example is `m/636-expose-new-temporary-metric(s)-on-when-pointer-table-is-used-as-fallback`, so i cannot do just
```
git checkout m/636-expose-new-temporary-metric(s)-on-when-pointer-table-is-used-as-fallback
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: Here's example what I'm talking about:
```
              val diff = b.diff(a)

              val actions = for {
                metrics <- metrics.toSeq
                log     <- log.toSeq
                topic   <- diff.toSeq
              } yield for {
                _ <- metrics.topicsFallback(topic)
                _ <- log.warn(s"`pointer` contains topic $topic while `pointer2` doesn't")
              } yield {}

              actions.sequence_.as(a ++ b)
```

PR #645, opened 2024-07-19
upgrade warnings to errors

mr-git:
19 places with `@nowarn` are still there, but those can be fixed only in next major release

### DIFF
--- a/cassandra/src/main/scala/com/evolutiongaming/kafka/journal/cassandra/CreateKeyspace.scala
+++ b/cassandra/src/main/scala/com/evolutiongaming/kafka/journal/cassandra/CreateKeyspace.scala
@@ -11,20 +11,20 @@ trait CreateKeyspace[F[_]] {
   def apply(config: KeyspaceConfig): F[Unit]
 }
 
-object CreateKeyspace { self =>
+object CreateKeyspace {
 
   def empty[F[_]: Applicative]: CreateKeyspace[F] = (_: KeyspaceConfig) => ().pure[F]
 
   def apply[F[_]: Monad: CassandraCluster: CassandraSession: LogOf]: CreateKeyspace[F] = new CreateKeyspace[F] {
 
-    def apply(config: KeyspaceConfig) = {
+    def apply(config: KeyspaceConfig): F[Unit] = {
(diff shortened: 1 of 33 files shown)


cassandra/src/main/scala/com/evolutiongaming/kafka/journal/cassandra/CreateKeyspace.scala
```diff
@@ -11,20 +11,20 @@ trait CreateKeyspace[F[_]] {
   def apply(config: KeyspaceConfig): F[Unit]
 }
 
-object CreateKeyspace { self =>
```
dfakhritdinov, line 14: IMO, its fine to keep `self` reference because Yaroslav used nested classes like `Main` in lots of places for simplifying navigation in IDE.

Its minor thing, feel free to ignore

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraTest.scala
```diff
@@ -444,7 +446,7 @@ object EventualCassandraTest {
   }
 
   def statementsOf(segmentNrsOf: SegmentNrsOf[StateT], segments: Segments): EventualCassandra.Statements[StateT] = {
-    implicit val concurrentStateT = ConcurrentOf.fromMonad[StateT]
+    val concurrentStateT: Concurrent[StateT] = ConcurrentOf.fromMonad[StateT]
```
dfakhritdinov, line 449: Why it was changed from been `implicit`?

dfakhritdinov (approved):

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandraTest.scala
mr-git: when type gets exposed (as newer Scala requires), the evaluation of value gets in loop as closest instance it can find is itself, IIUC

PR #650, opened 2024-08-12
Expose `set expire` API

dfakhritdinov:


journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedKeyJournal.scala
```diff
@@ -30,6 +30,11 @@ trait ReplicatedKeyJournal[F[_]] {
     offset: Offset,
     timestamp: Instant,
   ): F[Changed]
+
+  def expire(
```
migesok, line 34: I'd add a single sentence scaladoc explaining what does it do, smth like "Updates expiration time for a key". Otherwise looks legit

migesok: If you add, you are free to resolve and merge it yourself, I have no objections.

PR #657, opened 2024-08-28
DRAFT: Scala3 cross-compilation;

FunFunFine:


build.sbt
```diff
@@ -7,17 +7,35 @@ lazy val commonSettings = Seq(
   organizationHomepage := Some(url("https://evolution.com")),
   homepage := Some(url("https://github.com/evolution-gaming/kafka-journal")),
   startYear := Some(2018),
-  crossScalaVersions := Seq("2.13.14"),
+  crossScalaVersions := Seq("2.13.14", "3.3.3"),
```
mr-git, line 10: please update matrix in `ci.yml` file too

core/src/main/scala/com/evolutiongaming/kafka/journal/FromConfigReaderResult.scala
```diff
@@ -16,7 +18,7 @@ object FromConfigReaderResult {
 
   implicit def lift[F[_]: ApplicativeThrowable]: FromConfigReaderResult[F] = {
     new FromConfigReaderResult[F] {
-      def apply[A](a: ConfigReader.Result[A]) = {
+      def apply[A: ClassTag](a: ConfigReader.Result[A]): F[A] = {
```
mr-git, line 21: why we need `ClassTag` here?

core/src/main/scala/com/evolutiongaming/kafka/journal/HostName.scala
```diff
@@ -12,10 +12,68 @@ object HostName {
...
+import java.net.InetAddress
+import java.util.concurrent.Executors
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.sys.process.*
+import scala.util.Properties
+import scala.util.control.NonFatal
+
+/** Copied from https://github.com/evolution-gaming/hostname */
```
mr-git, line 32: is this temporary copy till `hostname` gets Scala 3 artifacts?

core/src/main/scala/com/evolutiongaming/kafka/journal/util/GracefulFiber.scala
```diff
@@ -23,10 +23,11 @@ object GracefulFiber {
...
             def cancel = {
-              for {
-                cancel <- cancelRef.getAndSet(true)
-                _      <- if (cancel) ().pure[F] else fiber.joinWithNever
-              } yield {}
+              cancelRef.getAndSet(true)
+                .flatMap(cancel =>
+                  (if (cancel) ().pure[F] else fiber.joinWithNever.void)
+
+                )
```
mr-git, line 30: why replace for comprehension with `.flatMap`?

I can see usage of `Concurrent[F].when(cancel) {fiber.joinWithNever}` as possible improvement, but why structural code change?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
```diff
@@ -317,7 +317,28 @@ object Journal {
...
+
+          pointerCur = objCur.atKeyOrUndefined("pointer")
+          pointer <- if (pointerCur.isUndefined) Right(CallTimeThresholds().pointer) else fdReader.from(pointerCur)
+
+          deleteCur = objCur.atKeyOrUndefined("delete")
+          delete <- if (deleteCur.isUndefined) Right(CallTimeThresholds().delete) else fdReader.from(deleteCur)
+
+          purgeCur <- objCur.atKeyOrUndefined("purge")
+          purge <- if (purgeCur.isUndefined) Right(CallTimeThresholds().purge) else fdReader.from(purgeCur)
+        } yield CallTimeThresholds(append, read, pointer, delete, purge)
```
mr-git, line 339: please elaborate why this is required

project/Dependencies.scala
```diff
@@ -5,24 +5,24 @@ object Dependencies {
...
-  val skafka                     = "com.evolutiongaming"    %% "skafka"                         % "16.0.0"
+  val random                     = "com.evolution"    %% "random"                         % "1.0.4"
+  val retry                      = "com.evolutiongaming"    %% "retry"                          % "3.1.0"
+  val sstream                    = "com.evolutiongaming"    %% "sstream"                        % "1.0.2"
+  val skafka                     = "com.evolutiongaming"    %% "skafka"                         % "16.4.0"
   val `akka-test-actor`          = "com.evolutiongaming"    %% "akka-test-actor"                % "0.1.0"
-  val scache                     = "com.evolution"          %% "scache"                         % "5.0.0"
-  val `resource-pool`            = "com.evolution"          %% "resource-pool"                  % "1.0.4"
+  val scache                     = "com.evolution"          %% "scache"                         % "5.1.2"
+  val `resource-pool`            = "com.evolution"          %% "resource-pool"                  % "1.0.5-SNAPSHOT"
```
mr-git, line 25: I suppose those `*-SNAPSHOT` dependencies are only for local development, right?

mr-git: please take in account that some libraries might bring in some transitively incompatible dependencies - please be careful

core/src/main/scala/com/evolutiongaming/kafka/journal/HostName.scala
FunFunFine: Both yes and no. 
Some of evo single-class libraries make almost no sense to me, so I decided to put this one here, especially since there's a wrapper for it anyway. 
Also I was not able to build the `hostname` easily, there are some weird problems with JVM and SBT or something like that.

If you feel like it should stay as a dependency, I'll update the library. But if you agree on it making no sense, then I'll refactor this file and make it nice.

core/src/main/scala/com/evolutiongaming/kafka/journal/FromConfigReaderResult.scala
FunFunFine: the version of pureconfig that was published for Scala 3, some of its APIs were updated and now require ClassTag

core/src/main/scala/com/evolutiongaming/kafka/journal/util/GracefulFiber.scala
FunFunFine: so I just did `Desugar for-comprehension` in IDEA and removed that last .map

FunFunFine: Perhaps it might compile with .void that I also put there, I'll check

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
FunFunFine: that's the DRAFT part. I'm not sure how to proceed with the config codec derivation since cross-compilation doesn't work with macros. This one is here for now to get the subproject to compile

project/Dependencies.scala
FunFunFine: yes, I have a tree of those dependencies written, there will be more MRs.

migesok:
reimplemented already and merged in 4.3.0

PR #658, opened 2024-08-29
remove reading of `pointer` table, but keep writing (#637)

### LINKED ISSUE #637: remove the fallbacks and release the version
Part of epic: #635

After confirming that `pointer` table is not used to get useful data as fallback, have to remove the fallback code to make sure that no client is reading `pointer` table any more.

Have to release and deploy new version across all applications.

mr-git:
closes #637


mr-git:
Should I make one more PR for 0-series (CE2)?

journal/src/main/scala/com/evolutiongaming/kafka/journal/eventual/ReplicatedJournal.scala
```diff
@@ -150,18 +150,6 @@ object ReplicatedJournal {
     def delete(topic: Topic, latency: FiniteDuration): F[Unit]
 
     def purge(topic: Topic, latency: FiniteDuration): F[Unit]
-
-    // TODO MR remove with next major release
```
dfakhritdinov, line 154: Are we doing major release already? Or the comment is wrong?

mr-git: wrong comment - at the time of writing I didn't have fully worked out the plan for rollout

mr-git: we could have left them deprecated for sake of binary compatibility, but all of them are our "internal API", which is public only because of the way, how Scala generates Java bytecode.

dfakhritdinov (approved):

PR #665, opened 2024-09-11
Add meta_record_id into journal & record_id metajournal tables for gaining correlation between them

dfakhritdinov:
### Problem
KJ can be in inconsistent state when (after purging) meta from `metajournal` was deleted while events from `journal` left behind. By itself its not an issue, but after the journal used again and new meta inserted - then old events will be available (probably with duplicated `seqNr`).

### Solution 
Create correlation between meta and events that belong to it, i. e. were created while the meta was present. Then, on reading journal, only events that correlate with meta must be used.

### Implementation 
Add `meta_record_id` column to `journal` and `record_id` column to `metajournal` tables. In appending events `RecordId` from meta will be used (if meta has it) and if meta is not present yet - new `RecordId` generated. On reading compare `RecordId` from meta and event ignoring events with another value. `RecordId` is optional value thus validation takes place only if its present in meta.

### DIFF
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
@@ -5,9 +5,10 @@ import cats.effect.syntax.all.*
 import cats.effect.{Concurrent, Resource}
 import cats.syntax.all.*
 import cats.{MonadThrow, Parallel}
-import com.evolutiongaming.catshelper.{LogOf, MeasureDuration, ToTry}
+import com.evolutiongaming.catshelper.{Log, LogOf, MeasureDuration, ToTry}
 import com.evolutiongaming.kafka.journal.Journal.DataIntegrityConfig
 import com.evolutiongaming.kafka.journal.eventual.*
+import com.evolutiongaming.kafka.journal.eventual.cassandra.JournalStatements.JournalRecord
 import com.evolutiongaming.kafka.journal.util.CatsHelper.*
 import com.evolutiongaming.kafka.journal.util.StreamHelper.*
 import com.evolutiongaming.kafka.journal.{cassandra as _, *}
@@ -122,7 +123,8 @@ object EventualCassandra {
       statements  <- Statements.of(schema, segmentNrsOf, Segments.default, consistencyConfig.read)
       _           <- log.info(s"kafka-journal version: ${Version.current.value}")
     } yield {
-      val journal = apply1[F](statements, dataIntegrity).withLog(log)
+      implicit val log1 = log
+      val journal       = apply2[F](statements, dataIntegrity).withLog(log)
       metrics
         .fold(journal) { metrics => journal.withMetrics(metrics) }
         .enhanceError
@@ -145,7 +147,13 @@ object EventualCassandra {
     * The implementation itself is abstracted from the calls to Cassandra which
     * should be passed as part of [[Statements]] parameter.
     */
+  @deprecated("Use apply2 instead", "3.6.0")
   def apply1[F[_]: MonadThrow](statements: Statements[F], dataIntegrity: DataIntegrityConfig): EventualJournal[F] = {
+    implicit val log = Log.empty[F]
+    apply2(statements, dataIntegrity)
+  }
+
+  def apply2[F[_]: MonadThrow: Log](statements: Statements[F], dataIntegrity: DataIntegrityConfig): EventualJournal[F] = {
 
     new Main with EventualJournal[F] {
 
@@ -166,12 +174,12 @@ object EventualCassandra {
               statement(key, segment.nr, range).map { record => (record, segment) }
             }
 
-            val events =
+            val records =
               read(from, Segment(from, head.segmentSize))
                 .chain {
                   case (record, segment) =>
                     for {
-                      from    <- record.seqNr.next[Option]
+                      from    <- record.event.seqNr.next[Option]
                       segment <- segment.next(from)
                     } yield {
                       read(from, segment)
@@ -179,25 +187,66 @@ object EventualCassandra {
                 }
                 .map { case (record, _) => record }
 
-            if (dataIntegrity.seqNrUniqueness) {
-              events
-                .stateful(from) {
-                  case (seqNr, record) =>
-                    if (seqNr <= record.seqNr) {
-                      val seqNr1 = record
-                        .seqNr
-                        .next[Option]
-                      (seqNr1, Stream[F].single(record))
-                    } else {
-                      val msg =
-                        s"Data integrity violated: seqNr $seqNr duplicated in multiple records from eventual journal, key $key"
-                      val err = new JournalError(msg)
-                      (seqNr.some, err.raiseError[F, EventRecord[EventualPayloadAndType]].toStream)
-                    }
+            val records1 = {
+              head.recordId.fold { records } { fromMeta =>
+                records.flatMap { event =>
+                  event.metaRecordId match {
+
+                    case None if dataIntegrity.correlateEventsWithMeta =>
+                      for {
+                        _ <- Log[F].error(s"Data integrity violated: event $event is orphan, key $key").toStream
+                        r <- Stream.empty[F, JournalRecord]
+                      } yield r
+
+                    case None =>
+                      Log[F]
+                        .warn(s"Disabled data integrity violated: event $event is orphan, key $key")
+                        .toStream
+                        .as(event)
+
+                    case Some(fromEvent) if fromEvent == fromMeta =>
+                      Stream.single(event)
+
+                    case Some(_) if dataIntegrity.correlateEventsWithMeta =>
+                      for {
+                        _ <- Log[F].error(s"Data integrity violated: event $event belongs to purged meta, key $key").toStream
+                        r <- Stream.empty[F, JournalRecord]
+                      } yield r
+
+                    case Some(_) =>
+                      Log[F]
+                        .warn(s"Disabled data integrity violated: event $event belongs to purged meta, key $key")
+                        .toStream
+                        .as(event)
+
+                  }
                 }
-            } else {
-              events
+              }
             }
+
+            val records2 =
+              if (dataIntegrity.seqNrUniqueness) {
+                records1
+                  .stateful(from) {
+                    case (seqNr, record) =>
+                      if (seqNr <= record.event.seqNr) {
+                        val seqNr1 = record
+                          .event
+                          .seqNr
+                          .next[Option]
+                        (seqNr1, Stream[F].single(record))
+                      } else {
+                        val msg =
+                          s"Data integrity violated: seqNr $seqNr duplicated in multiple records from eventual journal, key $key"
+                        val err = new JournalError(msg)
+                        (seqNr.some, err.raiseError[F, JournalRecord].toStream)
+                      }
+                  }
+              } else {
+                records1
+              }
+
+            records2.map(_.event)
           }
 
           head.deleteTo match {
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalHead.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalHead.scala
@@ -15,6 +15,14 @@ import com.evolutiongaming.scassandra.{DecodeRow, EncodeRow}
   * @param segmentSize the segmentSize used to write a journal to Cassandra
   * @param seqNr the sequence number of the last event in journal stored in Cassandra
   * @param deleteTo the last sequence number deleted from journal if any
+  * @param expiry the expiry time of the journal. On expiry the journal will be purged
+  * @param recordId unique ID of metadata entry, created once and _never_ updated. 
+  *   Used to correlate metadata with events inserted while metadata is actual: 
+  *   if purging journal will by mistake delete not all events, then after inserting new events
+  *   old ones will not be correlated with new metadata thus excluded from recovery.
+  *   The field is optional, because it was added later and old records do not have it,
+  *   thus while al least one record in `metajournal` table does not have `record_id` field - it should be optional.
+  *   `metajournal` record can be _created_ with `record_id` field, but never updated! 
   */
 final case class JournalHead(
   partitionOffset: PartitionOffset,
@@ -22,6 +30,7 @@ final case class JournalHead(
   seqNr: SeqNr,
   deleteTo: Option[DeleteTo] = none,
   expiry: Option[Expiry]     = none,
+  recordId: Option[RecordId] = none,
 )
 
 object JournalHead {
@@ -35,6 +44,7 @@ object JournalHead {
           seqNr           = row.decode[SeqNr],
           deleteTo        = row.decode[Option[DeleteTo]]("delete_to"),
           expiry          = row.decode[Option[Expiry]],
+          recordId        = row.decode[Option[RecordId]]("record_id"),
         )
       }
   }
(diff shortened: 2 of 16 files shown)


journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
```diff
@@ -332,17 +332,25 @@ object Journal {
...
-    def seqNrUniqueness: Boolean
-  }
+  final case class DataIntegrityConfig(
+    /**
+      * If true then duplicated [[SeqNr]] in events will cause [[JournalError]] `Data integrity violated`
+      */
+    seqNrUniqueness: Boolean,
+
+    /**
+      * If true then events with [[CorrelationId]] different from one in metadata will be filtered out
```
migesok, line 342: I think you missed this one in refactoring. I'd suggest to use text search on the project just to be sure.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.evolutiongaming.scassandra.*
+
+final case class RecordId(value: String) extends AnyVal {
+  override def toString: String = value
+}
+
+object RecordId {
+  def unsafe(value: String): RecordId = new RecordId(value)
+
+  def of[F[_]: Monad: UUIDGen]: F[RecordId] =
```
migesok, line 15: I'm not sure about the naming style in this project but I think `of` shouldn't result in new unique values generated - I'd renamed to smth like `generate` or `new` so it is visible in the code where new values are generated

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
```diff
@@ -0,0 +1,24 @@
...
+import cats.effect.std.UUIDGen
+import cats.syntax.all.*
+import com.evolutiongaming.scassandra.*
+
+final case class RecordId(value: String) extends AnyVal {
+  override def toString: String = value
+}
+
+object RecordId {
+  def unsafe(value: String): RecordId = new RecordId(value)
```
migesok, line 13: I don't think there is anything unsafe about this method - it appears to be redundant if you already have `apply`. I'd remove it.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalHead.scala
```diff
@@ -15,13 +15,19 @@ import com.evolutiongaming.scassandra.{DecodeRow, EncodeRow}
   * @param segmentSize the segmentSize used to write a journal to Cassandra
   * @param seqNr the sequence number of the last event in journal stored in Cassandra
   * @param deleteTo the last sequence number deleted from journal if any
+  * @param expiry the expiry time of the journal. On expiry the journal will be purged
+  * @param recordId unique ID of metadata entry, created once and _never_ updated.
```
migesok, line 19: I'd move the bulk of description to scaladoc of `RecordId` (check my comment there). Here it's better to describe in which cases `recordId` can be None - when we will look at the code 2 years from now, it might appear as if the field was always there and one might wonder what's the special meaning behind None case here.

migesok (changes requested):

mr-git:
let's adjust title of this PR

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
```diff
@@ -0,0 +1,24 @@
+package com.evolutiongaming.kafka.journal.eventual.cassandra
+
+import cats.Monad
+import cats.effect.std.UUIDGen
+import cats.syntax.all.*
+import com.evolutiongaming.scassandra.*
+
+final case class RecordId(value: String) extends AnyVal {
```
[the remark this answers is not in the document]
dfakhritdinov: I will more the description but honestly see no reason for having case class `MetaRecordId` because its used in "meta" record already

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
```diff
@@ -222,6 +230,11 @@ object JournalStatements {
                 metadata        = metadata,
                 headers         = headers,
               )
+
+              JournalRecord(
+                event        = eventRecord,
+                metaRecordId = row.decode[Option[RecordId]]("meta_record_id"),
+              )
```
mr-git, line 237: why we need this extra wrapper and why not put `metaRecordId` as field in `EventRecord`?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
```diff
@@ -31,6 +33,7 @@ object JournalStatements {
        |timestamp TIMESTAMP,
        |origin TEXT,
        |version TEXT,
+       |meta_record_id TEXT,
```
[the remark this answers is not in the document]
dfakhritdinov: No and not sure that I should, will check anyway

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
dfakhritdinov: I never have seen such notation within KJ thus no, I will keep `of`

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
dfakhritdinov: I'd keep it with its name. `unsafe` means that one can put arbitrary value into `RecordId` ignoring its rules

mr-git: and one can use `RecordId#apply` from non-private case class constructor directly ;)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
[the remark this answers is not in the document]
dfakhritdinov: done

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
dfakhritdinov: fixed, thanks

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalHead.scala
dfakhritdinov: I added explanation _why_ its optional and _when_ we can make it required field

persistence/src/test/resources/akka/persistence/kafka/journal/kafka-journal.conf
```diff
@@ -19,4 +19,5 @@ consumer-pool {
 }
 data-integrity {
   seq-nr-uniqueness =  true
+  correlate-events-with-meta = true
```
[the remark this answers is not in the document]
dfakhritdinov: Please check `DataIntegrityConfig` class

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
```diff
@@ -166,38 +174,79 @@ object EventualCassandra {
...
+
+                    case Some(_) if dataIntegrity.correlateEventsWithMeta =>
+                      for {
+                        _ <- Log[F].error(s"Data integrity violated: event $event belong to purged meta, key $key").toStream
+                        r <- Stream.empty[F, JournalRecord]
+                      } yield r
+
+                    case Some(_) =>
+                      Log[F]
+                        .warn(s"Disabled data integrity violated: event $event belong to purged meta, key $key")
```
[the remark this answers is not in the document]
dfakhritdinov: applied manually

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
```diff
@@ -166,38 +174,79 @@ object EventualCassandra {
...
+                        .warn(s"Disabled data integrity violated: event $event is orphan, key $key")
+                        .toStream
+                        .as(event)
+
+                    case Some(fromEvent) if fromEvent == fromMeta =>
+                      Stream.single(event)
+
+                    case Some(_) if dataIntegrity.correlateEventsWithMeta =>
+                      for {
+                        _ <- Log[F].error(s"Data integrity violated: event $event belong to purged meta, key $key").toStream
```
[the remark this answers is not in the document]
dfakhritdinov: applied manually

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
dfakhritdinov: Great question, I spend some time thinking what should I do.

`EventRecord` are created in `ReplicateRecords` in module `replicator` and passed down into `ReplicatedCassandra` in `cassandra` module. On level of `ReplicateRecords` I cannot provide instance of `RecordId` because metadata not yet available, it will be queried deeper in `cassandra` module because its purely Cassandra part, designed to overcome its limitations. On `replicator` level internal storage schema does not matter and its constraints should not leak out as well.

I choose having additional abstraction layer to avoid exposing `RecordId` in `replicator` (as unrelated to Replicator) and to avoid having case class with field that will be overrided later deeper in code.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/EventualCassandra.scala
```diff
@@ -166,38 +174,79 @@ object EventualCassandra {
...
+                    case (seqNr, record) =>
+                      if (seqNr <= record.event.seqNr) {
+                        val seqNr1 = record
+                          .event
+                          .seqNr
+                          .next[Option]
+                        (seqNr1, Stream[F].single(record))
+                      } else {
+                        val msg =
+                          s"Data integrity violated: seqNr $seqNr duplicated in multiple records from eventual journal, key $key"
```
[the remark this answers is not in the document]
dfakhritdinov: this message exists for a while, lets keep it

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/JournalStatements.scala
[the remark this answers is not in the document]
dfakhritdinov: `UUID` used as Cassandra field type for `RecordId`, thanks

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
```diff
@@ -14,21 +17,25 @@ import com.evolutiongaming.scassandra.*
   * that deleted record and recreated one are not considered the same while sharing
   * same private key.
   */
-final case class RecordId(value: String) extends AnyVal {
-  override def toString: String = value
+final case class RecordId(value: UUID) extends AnyVal {
+  override def toString: String = value.toString()
```
mr-git, line 21: in other places, where we want to disable default `apply` (to force usage of `of` or `unsafe`), we use `sealed abstract case class` instead of regular case class

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -38,7 +38,7 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
   private val origin          = Origin("origin")
   private val version         = Version.current
   private val uuid            = UUID.randomUUID
-  private val recordId        = RecordId.unsafe(uuid.toString)
+  private val recordId        = RecordId(uuid)
```
mr-git, line 41: here we use default `apply` in `unsafe` fashion, right?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala
dfakhritdinov: I think you used false assumption here: I do not see default `apply` as dangerous constructor and do not want disabling it. Mainly because creating new `RecordId` via default `apply` require pre-generated `UUID`, so unsafe action should be done _before_ `RecordId` creation.

In `SeqNr` Yaroslav disabled constructor because sequence numbers are applicable for arithmetic operations, which are dangerous. `RecordId`s have only one relation between each other - equality, thus protecting against mutation of `UUID` is pointless (each mutation gives unequal to original `UUID` while `SeqNr` arithmetics not so strait-forward).

My reasoning mentioned before still are not enough to keep the `apply` public, because its actually _not required_ to be public 😄 . Actual thing that stops me from following your suggestion is impossibility of making _abstract value class_: value class (one that extends `AnyVal`) is final (even if its not marked as `final`) thus it cannot be `abstract` - that's it.

Alternatively I can mark the constructor as private:
```
final case class RecordId private (value: UUID) extends AnyVal
``` 
but that makes Scala 3 compiler unhappy:
```
[error] <dev dir>/kafka-journal/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/RecordId.scala:20:18: access modifiers for `copy` method are copied from the case class constructor under Scala 3 (or with -Xsource-features:case-apply-copy-access)
[error] Scala 3 migration messages are issued as errors under -Xsource:3. Use -Wconf or @nowarn to demote them to warnings or suppress.
[error] Applicable -Wconf / @nowarn filters for this fatal warning: msg=<part of the message>, cat=scala3-migration, site=com.evolutiongaming.kafka.journal.eventual.cassandra.RecordId
[error] final case class RecordId private (value: UUID) extends AnyVal {
[error]                  ^
[error] one error found
``` 

I prefer keeping the `apply` publicly available, as its now.

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
dfakhritdinov: no, default `apply` is not unsafe because just creating new instance of case class **is not unsafe**. What is actually unsafe is `UUID.randomUUID`: its part of `RecordId.of: F[RecordId]` and `RecordId.unsafe: RecordId` where unsafe action hidden under effect or factory method explicitly tells you that its unsafe.

mr-git (approved):
overall LGTM

mr-git:
in which version we shall deliver this: on 3.x branch or with 4.0.0? (have to update `version.sbt` file)

dfakhritdinov:
@mr-git, you've asked

> in which version we shall deliver this: on 3.x branch or with 4.0.0? (have to update `version.sbt` file)

Adding new Cassandra table is not a breaking change thus keeping `3.x.x` series is fine.

PR #669, opened 2024-09-16
add DistributeJob doc

dfakhritdinov:


replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/DistributeJob.scala
```diff
@@ -16,6 +16,30 @@ import com.evolutiongaming.skafka.{Partition, Topic, TopicPartition}
 
 import scala.concurrent.duration.*
 
+/**
+ * [[DistributeJob]] is a service that distributes jobs across application nodes.
+ * 
+ * Distribution done with help of Apache Kafka consumers working within a single consumer group.
```
mr-git, line 22: Shall we mention that it is recommended to setup topic with higher than minimal number of partitions? 

For example, topic with 3 partitions will not distribute job across 13 nodes (10 nodes will be job-less) as topic with 39 partitions.

Shall we add docs also about `Distributed#single` here or on its definition?

dfakhritdinov: thanks for the advice, done

mr-git (approved):

PR #676, opened 2024-09-27
optimization of `append + delete` actions leads to inconsistent state in metajournal table

mr-git:
We optimize actions to reduce load on Cassandra when in single poll we get actions like:
* append @ 5
* append @ 6
* delete till 6

Optimization removed previous append actions, when `delete` would remove them after insertion.

The optimization keeps the `journal` table in shape, but it fails to update `metajournal` table, like when:
* `metajournal` table has entry with `deleteTo: 4; seqNr: 4`
* after receiving above-mentioned batch, we update `deleteTo` to `4` and `seqNr` stays as it was
* result is that we have `deleteTo: 4; seqNr: 4` (as `DeleteTo` got clamped down to existing `seqNr` to not overshoot future events in journal) - we lost updates: `seqNr: 6` and didn't update `deleteTo: 6`

This happens rarely, usually when client is quiet for some time and then issues "closing" statements. 

As this optimization drops `append @ 6`, it misses update of `expire_after` column in Cassandra.

### DIFF
--- a/eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
+++ b/eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
@@ -1498,6 +1498,9 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
       actual shouldEqual (expected, true).pure[Try]
     }
 
+    test(s"optimization: delete action sets head's expiration, $suffix") {
+      assert(false, "TODO MR implement!")
+    }
     test(s"optimization: `delete` action advances head's seq_nr, $suffix") {
       val id      = "id"
       val key     = Key(id = id, topic = topic0)
--- a/eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
+++ b/eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
@@ -1498,9 +1498,6 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
       actual shouldEqual (expected, true).pure[Try]
     }
 
-    test(s"optimization: delete action sets head's expiration, $suffix") {
-      assert(false, "TODO MR implement!")
-    }
     test(s"optimization: `delete` action advances head's seq_nr, $suffix") {
       val id      = "id"
       val key     = Key(id = id, topic = topic0)
--- a/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
+++ b/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
@@ -102,19 +102,11 @@ private[journal] object Batch {
 
               case (b: Delete, a: Action.Delete) =>
                 if (a.to > b.to) {
-                  if (tail.collectFirst { case b: Appends => cut(b, a) } getOrElse false) {
-                    val delete = deleteOf(a.to, origin orElse a.origin, version orElse a.version)
-                    delete :: Nil
-                  } else {
-                    val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
-                    delete :: tail
-                  }
-
-                  val cleanTail: List[Batch] = tail.map {
+                  val cleanTail: List[Batch] = tail.mapFilter {
                     case appends: Appends => dropDeleted(appends, a)
                     case delete: Delete   => delete.some
                     case purge: Purge     => purge.some
-                  }.flattenOption
+                  }
 
                   val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
                   delete :: cleanTail
--- a/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
+++ b/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
@@ -14,7 +14,9 @@ private[journal] object Batch {
 
   def of(records: Nel[ActionRecord[Action]]): List[Batch] = {
 
-    def cut(appends: Appends, delete: Action.Delete) = {
+    // returns `true` when we can optimize Cassandra usage by NOT inserting "previous" Append actions in DB
+    // because "current" Delete action will discard them all
+    def cut(appends: Appends, delete: Action.Delete): Boolean = {
       val append = appends.records.head.action
       append.range.to <= delete.to.value
     }
@@ -23,21 +25,23 @@ private[journal] object Batch {
       .foldLeft(List.empty[Batch]) { (bs, record) =>
         val offset = record.partitionOffset.offset
 
-        def appendsOf(records: Nel[ActionRecord[Action.Append]]) = {
+        def appendsOf(records: Nel[ActionRecord[Action.Append]]): Appends = {
           Appends(offset, records)
         }
 
-        def deleteOf(to: DeleteTo, origin: Option[Origin], version: Option[Version]) = {
+        def deleteOf(to: DeleteTo, origin: Option[Origin], version: Option[Version]): Delete = {
           Delete(offset, to, origin, version)
         }
 
-        def purgeOf(origin: Option[Origin], version: Option[Version]) = {
+        def purgeOf(origin: Option[Origin], version: Option[Version]): Purge = {
           Purge(offset, origin, version)
         }
 
-        def actionRecord[A <: Action](a: A) = record.copy(action = a)
+        def actionRecord[A <: Action](a: A): ActionRecord[A] = {
+          record.copy(action = a)
+        }
 
-        def origin = {
+        def origin: Option[Origin] = {
           bs.foldRight(none[Origin]) { (b, origin) =>
             origin orElse {
               b match {
@@ -49,7 +53,7 @@ private[journal] object Batch {
           }
         }
 
-        def version = {
+        def version: Option[Version] = {
           bs.foldRight(none[Version]) { (b, version) =>
             version orElse {
               b match {
@@ -128,9 +132,9 @@ private[journal] object Batch {
             }
         }
       }
-      .foldLeft(List.empty[Batch]) { (bs, b) =>
+      .foldLeft(List.empty[Batch]) { (bs, b) => // reverse order of batches
         b match {
-          case b: Appends => b.copy(records = b.records.reverse) :: bs
+          case b: Appends => b.copy(records = b.records.reverse) :: bs // reverse append actions
           case b: Delete  => b :: bs
           case b: Purge   => b :: bs
         }
--- a/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
+++ b/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
@@ -1,6 +1,6 @@
 package com.evolutiongaming.kafka.journal.replicator
 
-import cats.data.NonEmptyList as Nel
+import cats.data.NonEmptyList
 import cats.syntax.all.*
 import com.evolutiongaming.kafka.journal.*
 import com.evolutiongaming.skafka.Offset
@@ -12,7 +12,7 @@ private[journal] sealed abstract class Batch extends Product {
 
 private[journal] object Batch {
 
-  def of(records: Nel[ActionRecord[Action]]): List[Batch] = {
+  def of(records: NonEmptyList[ActionRecord[Action]]): List[Batch] = {
 
     // returns `true` when we can optimize Cassandra usage by NOT inserting "previous" Append actions in DB
     // because "current" Delete action will discard them all
@@ -21,11 +21,17 @@ private[journal] object Batch {
       append.range.to <= delete.to.value
     }
 
+    def dropDeleted(appends: Appends, delete: Action.Delete): Option[Appends] = {
+      // we cannot drop individual events in record as we do not parse the payload here (`records.head.action.payload: ByteVector`)
+      val validRecords = appends.records.collect { case record if record.action.range.to >= delete.to.value => record }
+      NonEmptyList.fromList(validRecords).map { Appends(appends.offset, _) }
+    }
+
     records
       .foldLeft(List.empty[Batch]) { (bs, record) =>
         val offset = record.partitionOffset.offset
 
-        def appendsOf(records: Nel[ActionRecord[Action.Append]]): Appends = {
+        def appendsOf(records: NonEmptyList[ActionRecord[Action.Append]]): Appends = {
           Appends(offset, records)
         }
 
@@ -78,7 +84,8 @@ private[journal] object Batch {
               case (b: Appends, a: Action.Delete) =>
                 if (cut(b, a)) {
                   val delete = deleteOf(a.to, origin orElse a.origin, version orElse a.version)
-                  delete :: Nil
+                  // preserve last `append` as we want to update `expireAfter` too, but we do not have [easy] access to it
+                  delete :: b.copy(records = NonEmptyList.of(b.records.head)) :: Nil
                 } else {
                   val delete = deleteOf(a.to, a.origin, a.version)
                   delete :: bs
@@ -88,7 +95,7 @@ private[journal] object Batch {
                 purgeOf(a.origin, a.version) :: Nil
 
               case (b: Delete, a: Action.Append) =>
-                appendsOf(Nel.of(actionRecord(a))) :: b :: tail
+                appendsOf(NonEmptyList.of(actionRecord(a))) :: b :: tail
 
               case (b: Delete, _: Action.Mark) =>
                 b.copy(offset = offset) :: tail
@@ -102,6 +109,15 @@ private[journal] object Batch {
                     val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
                     delete :: tail
                   }
+
+                  val cleanTail: List[Batch] = tail.map {
+                    case appends: Appends => dropDeleted(appends, a)
+                    case delete: Delete   => delete.some
+                    case purge: Purge     => purge.some
+                  }.flattenOption
+
+                  val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
+                  delete :: cleanTail
                 } else {
                   val delete = b.copy(offset = offset, origin = b.origin orElse a.origin, version = b.version orElse a.version)
                   delete :: tail
@@ -111,7 +127,7 @@ private[journal] object Batch {
                 purgeOf(a.origin, a.version) :: Nil
 
               case (b: Purge, a: Action.Append) =>
-                appendsOf(Nel.of(actionRecord(a))) :: b :: Nil
+                appendsOf(NonEmptyList.of(actionRecord(a))) :: b :: Nil
 
               case (b: Purge, _: Action.Mark) =>
                 b.copy(offset = offset) :: Nil
@@ -125,7 +141,7 @@ private[journal] object Batch {
 
           case Nil =>
             record.action match {
-              case a: Action.Append => appendsOf(Nel.of(actionRecord(a))) :: Nil
+              case a: Action.Append => appendsOf(NonEmptyList.of(actionRecord(a))) :: Nil
               case _: Action.Mark   => Nil
               case a: Action.Delete => deleteOf(a.to, a.origin, a.version) :: Nil
               case a: Action.Purge  => purgeOf(a.origin, a.version) :: Nil
@@ -143,7 +159,7 @@ private[journal] object Batch {
 
   final case class Appends(
     offset: Offset,
-    records: Nel[ActionRecord[Action.Append]],
+    records: NonEmptyList[ActionRecord[Action.Append]],
   ) extends Batch
 
   final case class Delete(
--- a/replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
+++ b/replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
@@ -60,7 +60,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
       ),
       (
         Nel.of(append(offset = 1, seqNr = 1), delete(offset = 2, seqNr = 1), append(offset = 3, seqNr = 2)),
-        List(deletes(offset = 2, seqNr = 1), appends(3, append(offset = 3, seqNr = 2))),
+        List(appends(1, append(offset = 1, seqNr = 1)), deletes(offset = 2, seqNr = 1), appends(3, append(offset = 3, seqNr = 2))),
       ),
       (
         Nel.of(
@@ -69,7 +69,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
           append(offset = 3, seqNr = 2),
           delete(offset = 4, seqNr = 2, origin = "origin2"),
         ),
-        List(deletes(offset = 4, seqNr = 2, origin = "origin1")),
+        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin1")),
       ),
       (
         Nel.of(
@@ -78,7 +78,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
           append(offset = 3, seqNr = 2),
           delete(offset = 4, seqNr = 2),
         ),
-        List(deletes(offset = 4, seqNr = 2, origin = "origin")),
+        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin")),
       ),
       (
         Nel.of(
@@ -87,7 +87,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
           delete(offset = 3, seqNr = 1, origin = "origin1"),
           delete(offset = 4, seqNr = 2, origin = "origin2"),
         ),
-        List(deletes(offset = 4, seqNr = 2, origin = "origin1")),
+        List(appends(2, append(offset = 2, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin1")),
       ),
       (
         Nel.of(
@@ -96,7 +96,10 @@ class BatchSpec extends AnyFunSuite with Matchers {
           delete(offset = 3, seqNr = 1),
           delete(offset = 4, seqNr = 2, origin = "origin"),
         ),
-        List(deletes(offset = 4, seqNr = 2, origin = "origin")),
+        List(
+          appends(offset = 2, append(offset = 2, seqNr = 2)),
+          deletes(offset = 4, seqNr = 2, origin = "origin"),
+        ),
       ),
       (Nel.of(delete(offset = 2, seqNr = 1), delete(offset = 3, seqNr = 2)), List(deletes(offset = 3, seqNr = 2))),
       (
@@ -121,7 +124,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
           delete(offset = 3, seqNr = 2),
           append(offset = 4, seqNr = 3),
         ),
-        List(deletes(offset = 3, seqNr = 2), appends(4, append(offset = 4, seqNr = 3))),
+        List(appends(2, append(offset = 2, seqNr = 2)), deletes(offset = 3, seqNr = 2), appends(4, append(offset = 4, seqNr = 3))),
       ),
       (
         Nel.of(
@@ -132,7 +135,7 @@ class BatchSpec extends AnyFunSuite with Matchers {
           delete(offset = 4, seqNr = 3),
           append(offset = 5, seqNr = 4),
         ),
-        List(deletes(offset = 4, seqNr = 3), appends(5, append(offset = 5, seqNr = 4))),
+        List(appends(3, append(offset = 3, seqNr = 3)),deletes(offset = 4, seqNr = 3), appends(5, append(offset = 5, seqNr = 4))),
       ),
       (
         Nel.of(
@@ -208,7 +211,6 @@ class BatchSpec extends AnyFunSuite with Matchers {
       ),
       (Nel.of(delete(offset = 0, seqNr = 1), purge(offset = 1)), List(purges(offset = 1))),
       (
-        // TODO MR can we optimize by dropping `delete` after `purge`?
         Nel.of(purge(offset = 0), delete(offset = 1, seqNr = 1)),
         List(purges(offset = 0), deletes(offset = 1, seqNr = 1)),
       ),
@@ -224,16 +226,8 @@ class BatchSpec extends AnyFunSuite with Matchers {
           delete(offset = 3, seqNr = 3),
           delete(offset = 4, seqNr = 5),
         ),
-        // TODO MR possible optimization for above:
-        //  * discard first 2 "appends"
-        //  * from third "append" filter out event with `seqNr: 5`
         List(
-          appends(
-            2,
-            append(offset = 0, seqNr = 1, seqNrs = 2),
-            append(offset = 1, seqNr = 3, seqNrs = 4),
-            append(offset = 2, seqNr = 5, seqNrs = 6),
-          ),
+          appends(offset = 2, append(offset = 2, seqNr = 5, seqNrs = 6)),
           deletes(offset = 4, seqNr = 5),
         ),
       ),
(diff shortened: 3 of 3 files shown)


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -429,21 +447,29 @@ private[journal] object ReplicatedCassandra {
...
+                                        (deleteTo.value, headSeqNr)
                                       }
-                                    } else if (deleteTo.value == seqNr) {
-                                      (seqNr, seqNr).pure[F]
+                                    } else if (deleteTo.value == headSeqNr) {
+                                      // all is fully pruned - nothing to do
+                                      (headSeqNr, headSeqNr).pure[F]
                                     } else {
+                                      // somehow `deleteTo` is ahead of head's `seqNr`
+                                      // TODO MR how this can happen? - in `def append` we clamp down `DeleteTo`
```
mr-git, line 458: I am not sure, how this can happen - my current guess is that due to some error we [used to?] get to unexpected state, possibly due to original bug, which gets fixed in this PR

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -429,21 +447,29 @@ private[journal] object ReplicatedCassandra {
...
                                           journalHead
                                             .copy(seqNr = seqNr1)
                                             .some
                                         }
                                       } yield {
-                                        (seqNr, seqNr1)
+                                        // assume that journal has been pruned till actual `deleteTo`
+                                        // and `seqNr` was not advanced for some reason
+//                                        (headSeqNr, seqNr1) // TODO MR do we need to prune this part of journal?
+                                        (seqNr1, seqNr1)
```
mr-git, line 472: This is an optimization to not issue useless `journal` cleanup query, but I am not sure, if we that is correct.

This part was incompletely covered with unit-tests originally, but if we assume that previous `delete` action was executed fully and correctly, then this should be OK and I can remove the `// TODO` line

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -455,13 +481,15 @@ private[journal] object ReplicatedCassandra {
...
-                                _ <- segmentNrs.parFoldMapA { segmentNr =>
-                                  statements
-                                    .deleteRecords(key, segmentNr)
-                                    .uncancelable
+                                _ <- Sync[F].whenA(from < to) {
+                                  segmentNrs.parFoldMapA { segmentNr =>
+                                    statements
+                                      .deleteRecords(key, segmentNr)
+                                      .uncancelable
+                                  }
```
mr-git, line 492: extra optimization to not issue delete statements when `from == to` - original unit-test had a unit-test, which kind of expected that no `delete` statements against `journal` table are done, while they were actually done, though not recorded in `StateT`

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1369,196 +1495,192 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+    StateT.unit { state =>
+      val k       = (key, segment)
+      val journal = state.journal
+      val entries = journal
+        .getOrElse(k, Map.empty)
+        .filter { case ((a, _), _) => a > seqNr }
+      val journal1 = if (entries.isEmpty) journal - k else journal.updated(k, entries)
+      state
+        .copy(journal = journal1)
+        .append(Action.DeleteRecordsTo(key, segment, seqNr))
```
mr-git, line 1525: this was missing

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1369,196 +1495,192 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
-    {
-      StateT.unit { state =>
-        val k = (key, segment)
-        state.copy(journal = state.journal - k)
-      }
+    StateT.unit { state =>
+      val k = (key, segment)
+      state
+        .copy(journal = state.journal - k)
+        .append(Action.DeleteRecords(key, segment))
```
mr-git, line 1534: this was missing

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1369,196 +1495,192 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+    StateT.unit { state =>
+      state
+        .updateMetaJournal(key, segment) { entry =>
           entry.copy(
             journalHead = entry.journalHead.copy(partitionOffset = partitionOffset, seqNr = seqNr, deleteTo = deleteTo.some),
             updated     = timestamp,
           )
         }
-      }
+        .append(Action.UpdateDeleteToAndSeqNr(key, segment, partitionOffset, timestamp, deleteTo, seqNr))
```
mr-git, line 1572: this was missing

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1369,196 +1495,192 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+          state.metaJournal.updated(k, entries1)
         }
-        state1
-          .getOrElse(state)
-          .append(Action.Delete(key, segment))
+        state.copy(metaJournal = metaJournal)
       }
+      state1
+        .getOrElse(state)
+        .append(Action.DeleteMetaJournal(key, segment))
```
mr-git, line 1661: this was missing

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1369,196 +1495,192 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
-        }
-        (state, offset)
+    StateT.success { state =>
+      val offset = for {
+        pointers <- state.pointers.get(topic)
+        pointer  <- pointers.get(partition)
+      } yield {
+        pointer.offset
       }
+      (state, offset)
```
mr-git, line 1683: should we record the any action here? probably have to migrate to `pointer2`

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1567,15 +1689,13 @@ object ReplicatedCassandraTest {
...
-          .updated(partition, entry)
-        state.copy(pointers = state.pointers.updated(topic, entries))
-      }
+    StateT.unit { state =>
+      val entry = PointerEntry(offset = offset, created = created, updated = updated)
+      val entries = state
+        .pointers
+        .getOrElse(topic, Map.empty)
+        .updated(partition, entry)
+      state.copy(pointers = state.pointers.updated(topic, entries))
```
mr-git, line 1698: should we record the any action here?

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1584,11 +1704,9 @@ object ReplicatedCassandraTest {
...
 
   val updatePointer: PointerStatements.Update[StateT] = { (topic, partition, offset, timestamp) =>
-    {
-      StateT.unit { state =>
-        state.updatePointer(topic, partition) { entry =>
-          entry.copy(offset = offset, updated = timestamp)
-        }
+    StateT.unit { state =>
+      state.updatePointer(topic, partition) { entry =>
+        entry.copy(offset = offset, updated = timestamp)
```
mr-git, line 1709: should we record the any action here?

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1598,11 +1716,9 @@ object ReplicatedCassandraTest {
...
 
   val selectTopics2: Pointer2Statements.SelectTopics[StateT] = { () =>
-    {
-      StateT.success { state =>
-        val topics = state.pointers.keySet.toSortedSet
-        (state, topics)
-      }
+    StateT.success { state =>
+      val topics = state.pointers.keySet.toSortedSet
+      (state, topics)
```
mr-git, line 1721: should we record the any action here?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -293,6 +293,7 @@ private[journal] object ReplicatedCassandra {
                         timestamp: Instant,
                         deleteTo: DeleteTo,
                         origin: Option[Origin],
+                        setSeqNr: Option[SeqNr],
```
dfakhritdinov, line 296: Could you please provide case when
```
setSeqNr.get > deleteTo.value // if setSeqNr is Some
```

mr-git: before this PR, we did clamp-down:
```scala
val deleteTo1 = deleteTo
  .value
  .min(seqNr1) // val seqNr1    = journalHead.deleteTo.fold { journalHead.seqNr } { _.value.max(journalHead.seqNr) }
  .toDeleteTo
```
in above snippet (inlined from original code), we clamp down `DeleteTo` to `min(seqNr, max(current_delete_to, seqNr))`

when "head" has `delete_to = 384` and `seqNr = 574` and we request `delete_to = 575`, we we were clamping it down to `574 = min(574, max(575, 574))`

mr-git: for new code, there is a unit-test case at https://github.com/evolution-gaming/kafka-journal/pull/676/files#diff-29aefcb72c7b9fba4c21d3237b491940e36217107a2b98c2ed8a3d2c4674c7b0R973 - we know that `Batch` would discard `append 575`, if in same poll we get `deleteTo 575` - `append 575` never happens, "head" stays at `seqNr 574`

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -311,42 +312,53 @@ private[journal] object ReplicatedCassandra {
...
+                              // clamp down to requested or max valid seqNr
+                              // in default protocol we have `Journal#delete(DeleteTo.max)`
+                              // `DeleteTo.max` can be used to discard all traces, but keep `seqNr`
+                              val validatedDeleteTo = deleteTo
                                 .value
-                                .min(seqNr1)
+                                .min(validatedSeqNr)
                                 .toDeleteTo
+
+                              val updateSeqNr = setSeqNr.isDefined
```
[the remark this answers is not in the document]
dfakhritdinov: And one more remark, more important than one above:

_Before the change_, code in `def delete` was also responsible for side-function - fix meta where
```
journalHead.deleteTo > journalHead.seqNr
```

Lets follow this cherry-picked lines _before the change_:
```
                              val deleteTo0 = journalHead.deleteTo
                              val seqNr       = journalHead.seqNr
                              val seqNr1     = deleteTo0.fold { seqNr } { _.value.max(seqNr) }
                              
                              if (seqNr1 == seqNr) { ... } else {
                                   update(seqNr1, ...)
                              }
```

As far as I see, _after the change_ `metajournal.seq_nr` will be updated only if 
```
setSeqNr.isDefined
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -311,42 +312,53 @@ private[journal] object ReplicatedCassandra {
...
+                              // different from value in head, must be less or equal to head's `seqNr`
+                              val updateDeleteTo = headDeleteTo.forall(_ < validatedDeleteTo)
```
dfakhritdinov, line 332: Before the change, `delete_to` was updated if new value does not equal to new one:
```
                                    if (deleteTo0.contains(deleteTo1)) {
                                      update()
                                    } else {
                                      update(deleteTo1)
                                    }
```
or
```
                                    if (deleteTo0.contains(deleteTo1)) {
                                      update(seqNr1)
                                    } else {
                                      update(seqNr1, deleteTo1)
                                    }
```
while you changed it on more loose condition:
```
headDeleteTo.forall(_ < validatedDeleteTo)
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: I see code for the same case in "main" part of `def delete(...)`:
```
val deleteTo0 = journalHead.deleteTo
val seqNr     = journalHead.seqNr
val seqNr1    = deleteTo0.fold { seqNr } { _.value.max(seqNr) }
if (seqNr1 == seqNr) { ... } else {
  update(seqNr1, ...)
}
```

New implementation lost it btw

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: I do not agree with the change here: if `deleteTo` is ahead of `seqNr` (both in meta) then KJ is in state:
- with inaccessible events in `journal` thus those events should be deleted
- with incorrect record in `metajournal` where `deleteTo > seqNr` thus it should be fixed

Please revert change on the line & cover it with tests (might be you already did)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: Please elaborate why events should not be deleted if `from == to`. I see it wrong because:

- purge is rare operation where optimizations not so important
- data inconsistency can lead to orphan events thus deleting them explicitly is fine

dfakhritdinov (changes requested):

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,7 +14,7 @@ private[journal] object Batch {
 
   def of(records: Nel[ActionRecord[Action]]): List[Batch] = {
 
-    def cut(appends: Appends, delete: Action.Delete) = {
+    def dropToBeDeletedAppends(appends: Appends, delete: Action.Delete): Boolean = {
```
dfakhritdinov, line 17: Sorry but I do not understand the name, might be docs will help? Actually _if only docs help_ then method name can stay as before

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: With state `(delete_to: 384, seqNr:574)` and action `DeleteTo(575)`, after stripping down original code we get:
```scala
val deleteTo0 = journalHead.deleteTo // 384
val seqNr     = journalHead.seqNr // 574
val seqNr1    = deleteTo0.fold { seqNr } { _.value.max(seqNr) } // 574 = max(384, 574)
val deleteTo1 = deleteTo.value.min(seqNr1).toDeleteTo // 574 = min(575, 574)

if (seqNr1 != seqNr && !deleteTo0.contains(deleteTo1)) { // 574 != 574 && 384 != 574
  update(seqNr1, deleteTo1) // doesn't happen
}
```

We would update both `seq_nr` and `delete_to` when both:
* head's `seq_nr` is not equal with head's `delete_to` (non-empty)
* and head's `delete_to` is not equal with clamped down requested action's `deleteTo`

---

In new code against the same state, we get:
```scala
val headDeleteTo = journalHead.deleteTo // 384
val headSeqNr    = journalHead.seqNr // 574
val validatedSeqNr: SeqNr = setSeqNr.fold { headSeqNr } { _.max(headSeqNr) } // 575 = max(575, 574)
val validatedDeleteTo = deleteTo.value.min(validatedSeqNr).toDeleteTo // 575 = min(575, 575)

val updateSeqNr = setSeqNr.isDefined // true
val updateDeleteTo = headDeleteTo.forall(_ < validatedDeleteTo) // true = 574 < 575

if (updateSeqNr && updateDeleteTo) { // true && true
  update(validatedSeqNr, validatedDeleteTo)
```

---

In old code there is no way to curry over information that `Delete` action had at least one `Append` action, which was discarded due to optimization

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: The assumption for this optimization is that previously `Delete` action was completed fully, like we have state `(delete_to: 575, seq_nr: 575)` - thus there is nothing expected in `journal` table (covered by unit-test `test(s"purge meta journal only...`, which doesn't expect any activity on `journal` table (based on assumption that `delete_to` was previously set correctly after corresponding events from `journal` were removed).

If we assume that `delete_to` could have been set wrongly and there are stale events in `journal` table, then on all `prune` actions we have to explicitly issue full cleanup of `journal` table - it is easily possible, though, depending on `segment_size`, can generate few to many queries (more for longer-lived entities).

mr-git: if we go with explicit cleanup, then we have to adjust unit-tests to expect activity on `journal` table too.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: BTW, purge is not rare - we run it every night to remove all expired journals - `PurgeExpired` finds all records on `metajournal` which have expired and then generates `Purge` action for each of them - every night there is a flood of `Purge` actions.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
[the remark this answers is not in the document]
mr-git: I am not sure that it is correct to "overload" the `Delete` action with extra functionality like:
> fix the case where somehow `delete_to > seq_nr` in `metajournal` table

It would mean that we do not trust our code to do the update correctly - I checked that in both client and replicator parts we clamp down `DeleteTo` to head's `seqNr`:
* in client: https://github.com/evolution-gaming/kafka-journal/blob/ba2031dd853b8f6d08b2db6d2e5e5d920cf9cfd3/journal/src/main/scala/com/evolutiongaming/kafka/journal/Journals.scala#L308
* in replicator: https://github.com/evolution-gaming/kafka-journal/blob/ba2031dd853b8f6d08b2db6d2e5e5d920cf9cfd3/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala#L309

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: Excuse me but your missed the question:

> // somehow `deleteTo` is ahead of head's `seqNr`

> I see code for the same case in "main" part of def delete(...):

> New implementation lost it btw

Thus the case is: initial state - `delete_to = 400` and `seq_nr = 300` when incoming `DeleteTo = 200`, then
```
                              val deleteTo0 = journalHead.deleteTo // 400
                              val seqNr     = journalHead.seqNr    // 300
                              val seqNr1    = deleteTo0.fold { seqNr } { _.value.max(seqNr) } // 400
                              val deleteTo1 = deleteTo // 200
                                .value
                                .min(seqNr1)
                                .toDeleteTo
                                  if (seqNr1 == seqNr) { // 400 != 300, ie false
                                    if (deleteTo0.contains(deleteTo1)) {
                                      ...
                                  } else {
                                     ... // update seq_nr, this is lost in your new implementation 
                                  }
```

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: > thus there is nothing expected in journal table

its wrong assumption because now (probably) some events are left behind and purging such journal will make them orphan. Cleaning orphan events are long and heavy job thus we will not run it often. I see no way how I can approve the PR with such "optimization"

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: how come this change is "fixing test"? I see it as fixing code because there's some test ... Please revert it

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
```diff
@@ -19,130 +19,15 @@ class BatchSpec extends AnyFunSuite with Matchers {
 
   for {
     (values, expected) <- List[(Nel[A], List[Batch])](
-      (Nel.of(mark(offset = 0)), Nil),
```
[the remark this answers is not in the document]
mr-git: thank you! now there are 41 cases

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: Unwrapped "before":
```scala
if ( ! journalHead.deleteTo.contains(deleteTo.value.min(journalHead.seqNr).toDeleteTo) ) {
// update `delete_to` fiels
}
```

Like "if head's delete_to is empty or doesn't contain min(requested or head's seq_nr)".

Unwrapped in this PR:
```scala
val s = setSeqNr.fold { journalHead.seqNr } { _.max(journalHead.seqNr) }
if (journalHead.deleteTo.forall(_ < deleteTo.value.min(s).toDeleteTo)) {
// update `delete_to` fiels
}
```

Now "if head's delete_to is empty of the existing value is smaller than min(requested or max(head's seq_nr or <from removed append action> ))"

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: I created the unit-test-case in `ReplicatedCassandraTest` like:
```scala
    test(s"fix delete_to, $suffix") {
      val id      = "id"
      val key     = Key(id = id, topic = topic0)
      val segment = segmentOfId(key)
      val stateT = journal.delete(
        key = key,
        Partition.min,
        Offset.unsafe(2),
        timestamp = timestamp1,
        deleteTo  = SeqNr.unsafe(200).toDeleteTo,
        origin    = origin.some,
      )

      val initial = State
        .empty
        .copy(
          metaJournal = Map(
            (
              (topic0, segment),
              Map(
                (
                  id,
                  MetaJournalEntry(
                    journalHead = JournalHead(
                      partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(1)),
                      segmentSize     = segmentSize,
                      seqNr           = SeqNr.unsafe(300),
                      deleteTo        = SeqNr.unsafe(400).toDeleteTo.some,
                      expiry          = none,
                      recordId        = none,
                    ),
                    created = timestamp0,
                    updated = timestamp0,
                    origin  = origin.some,
                  ),
                ),
              ),
            ),
          ),
        )

      val expected = State(
        metaJournal = Map(
          (
            (topic0, segment),
            Map(
              (
                id,
                MetaJournalEntry(
                  journalHead = JournalHead(
                    partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(2)),
                    segmentSize     = segmentSize,
                    seqNr           = SeqNr.unsafe(200), // <========= here test fails - actual value is `400`
                    deleteTo        = SeqNr.unsafe(200).toDeleteTo.some,
                    expiry          = none,
                    recordId        = none,
                  ),
                  created = timestamp0,
                  updated = timestamp1,
                  origin  = origin.some,
                ),
              ),
            ),
          ),
        ),
      )

      val actual = stateT.run(initial)
      actual shouldEqual (expected, true).pure[Try]
    }
```

It fails with

mr-git: we we "fix" by advancing `seqNr`... will add test-case to this PR...

dfakhritdinov: > If I read it correct, then journal table would be empty and on recover actor would start with seqNr: 300.

the state can be:
```
meta.seq_nr     = 300
meta.delete_to = 400

journal contains events from 0 to 400
```

> Is the setup you describe even possible?

I have no idea how this state is possible but:
1. i don't believe that the fix logic was added just for fun
2. i don't have any guarantee that we don't have such journals
3. the code _now_ does not allow it but could allow _in past_

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
dfakhritdinov: could you please explain why before `delete_to` would be updated even if 
```
deleteTo (incoming from an app) > meta.seq_nr > meta.delete_to
```

mr-git:
up to https://github.com/evolution-gaming/kafka-journal/pull/676/commits/01ea06203365bdb6ce1d8eab0afae2e8032114d7 there is "version 1" - which should be safe and has minimal changes in `Batch.scala`

for improved visibility, I can split PR in two: one with minimal changes and other with reworked `Batch.scala` - should I?

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -98,6 +109,15 @@ private[journal] object Batch {
                     val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
                     delete :: tail
                   }
+
+                  val cleanTail: List[Batch] = tail.map {
+                    case appends: Appends => dropDeleted(appends, a)
+                    case delete: Delete   => delete.some
```
rtar, line 115: I wonder if we really need previous deletes. :thinking:

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -98,6 +109,15 @@ private[journal] object Batch {
...
+                    case purge: Purge     => purge.some
+                  }.flattenOption
```
rtar, line 117: I think one can use `mapFilter` instead of `map+flattenOption` here.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -98,6 +109,15 @@ private[journal] object Batch {
...
+
+                  val delete = deleteOf(a.to, b.origin orElse a.origin, b.version orElse a.version)
```
rtar, line 119: I find it a bit awkward that we do the following:
```scala
if (a.to > b.to) {
  ..., b.origin orElse a.origin, b.version orElse a.version...
} else {
  ..., b.origin orElse a.origin, b.version orElse a.version...
}
```
If I understood it correctly, it means that sometimes the older origin/version will be used, and sometimes a newer one.

It does not matter, probably, and was like that even before this PR, but I just wanted to comment on it.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -74,7 +84,8 @@ private[journal] object Batch {
               case (b: Appends, a: Action.Delete) =>
                 if (cut(b, a)) {
                   val delete = deleteOf(a.to, origin orElse a.origin, version orElse a.version)
-                  delete :: Nil
+                  // preserve last `append` as we want to update `expireAfter` too, but we do not have [easy] access to it
+                  delete :: b.copy(records = NonEmptyList.of(b.records.head)) :: Nil
```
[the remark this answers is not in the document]
mr-git: yes, it is reversed.

the last test-case covers this case:
```scala
      (
        Nel.of( // state: delete_to = 384, seqNr = 573 (in Cassandra)
          append(offset = 1797039, seqNr = 574),
          append(offset = 1801629, seqNr = 575),
          mark(offset   = 1801632),
          delete(offset = 1801642, seqNr = 575),
        ),
        List(
          appends(offset = 1801632, append(offset = 1801629, seqNr = 575)),
          deletes(offset = 1801642, seqNr = 575),
        ),
      ),
```
From 2 `append` actions we keep the most recent one (`575`)

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: the logic os very convoluted here - idea is that `delete` could be the first action in stream, which "inits" journal - and we want (have?) to preserve the `origin` and `version` of journal's first event - there are several test-cases which test this for `origin` (but not for `version`).

I am not sure, why we need that, but maybe idea was that it would allow us to (just guessing):
* `origin` - find the pod, which created the journal
* `version` - provide an ability to make an educated guess on how many old journals we have - like if we can drop backwards compatibility for `kafka-journal` specific data

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: I do not see point in keeping and applying them, but it would be very challenging to remove them in current code - I've reworked it in #687

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: indeed, thank you!

rtar (approved):

PR #679, opened 2024-09-30
expose observed journal expirations as metrics

mr-git:
Currently there is no easy way to observer what expirations are used by different clients - only options are:
* peek at Kafka messages
* crawl the `journal` table
* search in source

With this change, `replicator` will expose new metric `*_kafka_journal_client_journal_expiration_info` which will expose [a sample] of expirations as observed in replicated data.

### DIFF
--- a/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
+++ b/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
@@ -72,11 +72,13 @@ private[journal] object ReplicateRecords {
           def measure(events: Nel[EventRecord[EventualPayloadAndType]], expireAfter: Option[ExpireAfter]) = {
             for {
               measurements <- measurements(records.size)
-              version       = events.last.version.map(_.value).getOrElse("none")
+              version       = events.head.version.map(_.value).getOrElse("none")
+              expiration    = events.head.metadata.payload.expireAfter.map(_.value.toString).getOrElse("none")
               result <- metrics.append(
                 events        = events.length,
                 bytes         = bytes,
                 clientVersion = version,
+                expiration    = expiration,
                 measurements  = measurements,
               )
               _ <- log.info(msg(events, measurements.replicationLatency, expireAfter))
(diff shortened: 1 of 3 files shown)


replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
```diff
@@ -72,11 +72,13 @@ private[journal] object ReplicateRecords {
           def measure(events: Nel[EventRecord[EventualPayloadAndType]], expireAfter: Option[ExpireAfter]) = {
             for {
               measurements <- measurements(records.size)
-              version       = events.last.version.map(_.value).getOrElse("none")
+              version       = events.head.version.map(_.value).getOrElse("none")
```
dfakhritdinov, line 75: Please elaborate why `events.last` was changed on `events.head`?

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
```diff
@@ -72,11 +72,13 @@ private[journal] object ReplicateRecords {
...
+              expiration    = events.head.metadata.payload.expireAfter.map(_.value.toString).getOrElse("none")
```
dfakhritdinov, line 76: I think using _first_ event is not accurate because KJ will take expiration for _last_ event as final.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
mr-git: for `version` it plays no role, but for `expiration` it might in scenarios when journal gets several different values depending on state.

my idea was that with this approach we would get wider (and actual) list of values, though as this happens on per-poll and Kafka might batch in different ways, it should not make much difference in long-term

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/ReplicateRecords.scala
mr-git: yeah, and cutting of polls can happen at any place - so eventually we will cover all values also with `last`, but dominating values will come first - I will change

dfakhritdinov (approved):

PR #682, opened 2024-10-01
add more test-cases for `ReplicatedCassandra`

mr-git:
required for better coverage before dealing with #676

### DIFF
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
@@ -424,7 +424,7 @@ private[journal] object ReplicatedCassandra {
                                       }
                                     } else if (deleteTo.value == seqNr) {
                                       (seqNr, seqNr).pure[F]
-                                    } else {
+                                    } else { // deleteTo.value > seqNr
                                       val seqNr1 = deleteTo.value
                                       for {
                                         _ <- update(seqNr1)
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
@@ -335,7 +335,11 @@ private[journal] object ReplicatedCassandra {
                                       for {
                                         segmentNrs <- from
                                           .toJournalSegmentNr(segmentSize)
-                                          .to[F] { to.toJournalSegmentNr(segmentSize) }
+                                          .to[F] {
+                                            // TODO MR why we don't use `to.next[Option].getOrElse(SegmentNr.max)` as in `def purge`?
+                                            //  see test-case `repeat purge again for the same offset` ???
+                                            to.toJournalSegmentNr(segmentSize)
+                                          }
                                         result <- {
                                           if (to >= seqNr) {
                                             segmentNrs.parFoldMapA { segmentNr =>
@@ -444,10 +448,15 @@ private[journal] object ReplicatedCassandra {
                                   .getOrElse { from }
                                   .toJournalSegmentNr(segmentSize)
                                   .to[F] {
+                                    // TODO MR for context see test-case `repeat purge again for the same offset`
                                     to
-                                      .next[Option]
+                                      .next[Option] // TODO MR why overshoot to possibly next segment?
                                       .getOrElse { journalHead.seqNr }
                                       .toJournalSegmentNr(segmentSize)
+                                    // TODO MR alternative, if overshoot is required
+//                                    to.toSegmentNr(segmentSize).next[Option].getOrElse(SegmentNr.max)
+                                    // TODO MR alternative, if same as in `def delete`
+//                                    to.toSegmentNr(segmentSize)
                                   }
                                 _ <- segmentNrs.parFoldMapA { segmentNr =>
                                   statements
--- a/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
+++ b/eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
@@ -335,11 +335,7 @@ private[journal] object ReplicatedCassandra {
                                       for {
                                         segmentNrs <- from
                                           .toJournalSegmentNr(segmentSize)
-                                          .to[F] {
-                                            // TODO MR why we don't use `to.next[Option].getOrElse(SegmentNr.max)` as in `def purge`?
-                                            //  see test-case `repeat purge again for the same offset` ???
-                                            to.toJournalSegmentNr(segmentSize)
-                                          }
+                                          .to[F] { to.toJournalSegmentNr(segmentSize) }
                                         result <- {
                                           if (to >= seqNr) {
                                             segmentNrs.parFoldMapA { segmentNr =>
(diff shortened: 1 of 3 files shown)


eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -334,7 +334,11 @@ private[journal] object ReplicatedCassandra {
                                       for {
                                         segmentNrs <- from
                                           .toSegmentNr(segmentSize)
-                                          .to[F] { to.toSegmentNr(segmentSize) }
+                                          .to[F] {
+                                            // TODO MR why we don't use `to.next[Option].getOrElse(SegmentNr.max)` as in `def purge`?
```
dfakhritdinov, line 338: here `to` is `SeqNr` not `SegmentNr`.

The deletion should happened to exactly `to` rather next sequence number after `to`

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -443,10 +447,15 @@ private[journal] object ReplicatedCassandra {
                                   .getOrElse { from }
                                   .toSegmentNr(segmentSize)
                                   .to[F] {
+                                    // TODO MR for context see test-case `repeat purge again for the same offset`
                                     to
-                                      .next[Option]
+                                      .next[Option] // TODO MR why overshoot to possibly next segment?
```
dfakhritdinov, line 452: Safety measure?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: the full code is `to.toSegmentNr(segmentSize)`

my question is related to how `purge` works - it overshoots by one segment

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: why here we do "safety measure", but not in `def delete` - they both do the same, don't they?

do you think that this is "by design" and not a bug?

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -443,10 +447,15 @@ private[journal] object ReplicatedCassandra {
...
                                       .getOrElse { journalHead.seqNr }
                                       .toSegmentNr(segmentSize)
+                                    // TODO MR alternative, if overshoot is required
```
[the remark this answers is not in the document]
mr-git: I do not understand why during purge we want to delete also next segment (when `to` is very last `seqNr` in `Segment`)

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
mr-git: for me it looks like `off by 1` error

dfakhritdinov: I'm afraid but I have no idea _why_ its done like that. On other hand I cannot accept dropping this code without understanding _why_ it was written and _what_ are consequences of a change.

eventual-cassandra/src/main/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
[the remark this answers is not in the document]
dfakhritdinov: Sorry but I prefer to "play safe" and change/delete things that I understand rather opposite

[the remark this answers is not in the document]
mr-git: It looks like work-around-compensation for #676, and we can keep it - it is just one extra CQL statement, which, most likely, will affect zero rows.

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -104,6 +104,7 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
 
       val expected = State(
         actions = List(
+          Action.DeleteRecords(key, SegmentNr.journal(SeqNr.unsafe(1), segmentSize)),
```
dfakhritdinov, line 107: Please elaborate why this action happened now but not before the change

mr-git: It was never recorded before - I added it here: 
https://github.com/evolution-gaming/kafka-journal/pull/682/files/32e7a11459e1c8720988033054639e53ce19ee64#diff-29aefcb72c7b9fba4c21d3237b491940e36217107a2b98c2ed8a3d2c4674c7b0R1885

Similar for deletion of record in `metajournal` table

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -890,7 +891,9 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
           Offset.unsafe(1),
           timestamp   = timestamp0,
           expireAfter = none,
-          events = Nel.of(eventRecordOf(seqNr = SeqNr.min, partitionOffset = PartitionOffset(Partition.min, Offset.min)).event),
+          events = Nel.of(
+            eventRecordOf(seqNr = SeqNr.min, partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(1))).event,
```
dfakhritdinov, line 895: Please elaborate why `Offset.min` was changed on `Offset.unsafe(1)`

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1086,89 +1092,219 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
-            SeqNr.unsafe(2).toDeleteTo,
+        val stateT = journal.purge(key, partitionOffset.partition, offset, timestamp1)
+
+        val expected = State(
+          actions = List(
+            Action.DeleteMetaJournal(key, segment),
+          ) ++ {
+            for {
+              i <- segments to (0, -1)
+            } yield Action.DeleteRecords(key, SegmentNr.of[Option](i.toLong).get)
```
dfakhritdinov, line 1193: If you need pure computation while some `F[_]` is used - try `Id` from cats, no potential runtime exceptions like in `Option#get`

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
mr-git: to make it obvious what value is used and to make it consistent with other lines where `Offset.unsafe(<other value>)` gets used. It doesn't carry any meaning in this test case.

dfakhritdinov (approved):

eventual-cassandra/src/test/scala/com/evolutiongaming/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
mr-git: I, shamelessly, copied it from other place without thinking. Adjusted, please recheck.

PR #687, opened 2024-10-16
rework algorithm in `Batch` to reduce variability and increases readability 

mr-git:
Rework logic in `Batch` as next step for #676 

Idea is as fallows:
* drop all actions before `purge`
* keep single `delete` action - the most recent one
* drop would be discarded `append` actions
* batch non-discarded `append` actions
* apply them in order: `purge`, `appends`, `delete`


replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,147 +22,79 @@ private[journal] object Batch {
...
-        }
-
-        def deleteOf(to: DeleteTo, origin: Option[Origin], version: Option[Version]): Delete = {
-          Delete(offset, to, origin, version)
-        }
-
-        def purgeOf(origin: Option[Origin], version: Option[Version]): Purge = {
-          Purge(offset, origin, version)
+        // collect `Append` and `Delete` actions
+        case (acc, record) => acc :+ record
```
rtar, line 34: I do not know if this is important, but this append will take `O(n)` for a list of size `n`.

I.e. the whole filtering loop will take `O(n)`. I.e. if the batch contains `100` events then there will be `1000` appends.

As far as I remember these batches are much smaller, so this could be a premature optimization from my side, but might be worth to double check.

mr-git: yeah, I was thinking about benchmarking existing and new code, but didn't do that, yet.

in `ReplicateRecords` we do few other traversals over all records from `poll`, if I am reading code correctly, then:
* in `TopicReplicator:116` we get all `records` from Kafka's `poll` result
* on `TopicReplicator:120` we group them in to `SortedMap` them by `partition`
* on `TopicReplicator:145` we group them by `key`
* on `TopicReplicator:164` we feed all those per-key records to `ReplicateRecords`
* on `ReplicateRecords:172` we filter them
* on `ReplicateRecords:175` we pass them to `ReplicateRecords:161` where we pass them to `Batch` 

in sum, before `Batch`, we have traversed them all at least 3 times already, though in `Batch` we will traverse them 3 additional times 😞

mr-git: maybe it is not so bad - in current version:
* we process all records with `foldLeft`
* if there are `delete` actions, we backtrack
* then at `Batch:143` we reverse batched items

in new version:
* we do `foldLeft` and `groupBy`
* then we traverse 3 groups

looks like we will do one more traverse, though it is hard to compare, as new version is more aggresisve with batching of `delete` actions (which are rare)

performance-wise this change affects only `replicator` part - easy to allocate more resources, if required.

consistency-wise new code feels better.

comprehension-wise - at least for me - new code looks easier to read and understand, it doesn't have backtracking in few of 12 match-cases and there is no reverse-logic

rtar: Doing more traverses is fine, because it just increments a constant, i.e. we keep it `O(n)`.

But having `:+` might lead to `O(n^2)`, which could be a problem for larger bets.

mr-git: checked on Live - according to metrics, which are scraped once a minute, on average across any minute we have batches with less than 1.2 records.

like it could be that we have 99 batches with size 1 and 1 batch with 20 records in it... over a minute we might have some really large batches, though realistically we rarely produce more than few records at time per aggregate.

rtar: I am thinking about the theoretical situation, when there is only 1 record per batch, normally, and then, suddenly (because of the glitch on producer side), 10 000 records come in. Then we will end up with 100 000 000 cycles instead of 10 000 x C (where C is some constant).

Or the batch size is strictly limited, anyway?

mr-git: by default - max is 500 (default in Kafka's config [max-poll-records](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#max-poll-records)).

I could:
* run performance tests with this PR vs "no-op" `Batch`-ing (the one, which doesn't batch anything)
* rework the implementation to use prepend and reverse at the end
* reimplement code to use `Vector` or `ListBuffer` instead of `List` - those have better append performance

mr-git: so, `original` is faster, but it does less reductions on edge cases - leads to execution of more Cassandra statements

mr-git: the latest version:
```
[info] Benchmark                            Mode  Cnt       Score      Error  Units
[info] BatchBenchmark.alternativeReversed  thrpt    5  117159.312 ±  918.479  ops/s
[info] BatchBenchmark.original             thrpt    5  126303.502 ± 3433.752  ops/s
```

@rtar, what do you think about new approach with internal collection state? performance wise it is the same, and all unit-tests pass too

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,139 +24,87 @@ private[journal] object Batch {
...
 
-          case Nil =>
-            record.action match {
-              case a: Action.Append => appendsOf(NonEmptyList.of(actionRecord(a))) :: Nil
-              case _: Action.Mark   => Nil
-              case a: Action.Delete => deleteOf(a.to, a.origin, a.version) :: Nil
-              case a: Action.Purge  => purgeOf(a.origin, a.version) :: Nil
-            }
+          case None =>
+            None
```
rtar, line 95: This usually means that the underlying `match...case` could be replaced by a simple `flatMap`.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,139 +24,83 @@ private[journal] object Batch {
...
+            Delete(po.offset, delete.to, delete.origin, delete.version)
         }
-
-        def deleteOf(to: DeleteTo, origin: Option[Origin], version: Option[Version]): Delete = {
-          Delete(offset, to, origin, version)
+        state.copy(
+          delete = delete_.some,
+        )
+
+      case (state, ActionRecord(append: Action.Append, po)) if state.delete.forall(_.to.value <= append.range.to) =>
```
rtar, line 54: I wonder if these `Delete` actions could come in a wrong order? Or that cannot happen?

I.e. like this:
```
Delete(3)
Append(1)
Append(2)
Append(3)
```
In this case, after `records.reverse` that `Delete(3)` will be handled last, and the appends will not be cleaned up.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,139 +24,83 @@ private[journal] object Batch {
...
-      NonEmptyList.fromList(validRecords).map { Appends(appends.offset, _) }
-    }
-
-    records
-      .foldLeft(List.empty[Batch]) { (bs, record) =>
-        val offset = record.partitionOffset.offset
-
-        def appendsOf(records: NonEmptyList[ActionRecord[Action.Append]]): Appends = {
-          Appends(offset, records)
+    val state = records.reverse.foldLeft(State()) {
```
rtar, line 27: Style comment: I would consider moving these `case .. =>` statements to `State` case class as methods. :blush:

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: passing test-case:
```scala
      (
        Nel.of(
          delete(offset = 1, to = 3),
          append(offset = 2, seqNr = 1),
          append(offset = 3, seqNr = 2),
          append(offset = 4, seqNr = 3),
        ),
        List(
          appends(offset = 4, append(offset = 2, seqNr = 1), append(offset = 3, seqNr = 2), append(offset = 4, seqNr = 3)),
          deletes(offset = 4, to = 3),
        ),
      ),
```
it would add and delete those events - sadly, they are not removed by `Batch`.

in original code, to pass test-case, the check should be:
```scala
        List(
          deletes(offset = 1, to = 3),
          appends(offset = 4, append(offset = 2, seqNr = 1), append(offset = 3, seqNr = 2), append(offset = 4, seqNr = 3)),
        ),
```

indeed, this creates inconsistency with original handling, as `Delete(3)` attempts to delete "in future" and in original version the events are "leaked" in `journal` table, while in this PR they will get added and removed.

if it would get applied in 2 batches - `Delete(3)` and then all `Append(*)`, then:
* in `metajournal` we would create head with `delete_to=3, seqNr = 3` (this is valid way to create new journal)
* all `Append(*)` would get ignored - those events would never become available for journal

when we apply them all in single batch, then:
* in `metajournal` we expect to create head with `delete_to=3, seqNr = 3`
* there should not be events in `journal` table - here the original code "leaks" them, while new one will cleanup

overall - such "delete in future" is not expected, though with bug in code it is possible. I am not sure, if there is an easy way how to overcome it - there is no way for `kafka-journal` to provide feedback to client about inconsistent or unexpected behaviour. 

test-case in original code for `ReplicatedCassandraTest` demonstrates that events are "leaked" (on recovery they will be ignored):
```scala
    test(s"batched `[delete(2), append(1), append(2)]` must update tables correctly, $suffix") {
      val id          = "id"
      val key         = Key(id = id, topic = topic0)
      val metaSegment = segmentOfId.metaJournal(key)

      val stateT = for {
        _ <- journal.delete(
          key       = key,
          partition = Partition.min,
          offset    = Offset.unsafe(1),
          timestamp = timestamp0,
          deleteTo  = SeqNr.unsafe(2).toDeleteTo,
          origin    = none,
        )
        _ <- journal.append(
          key         = key,
          partition   = Partition.min,
          offset      = Offset.unsafe(2),
          timestamp   = timestamp1,
          expireAfter = none,
          events = Nel.of(
            eventRecordOf(seqNr = SeqNr.unsafe(1), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(2))).event,
          ),
        )
        _ <- journal.append(
          key         = key,
          partition   = Partition.min,
          offset      = Offset.unsafe(3),
          timestamp   = timestamp1,
          expireAfter = none,
          events = Nel.of(
            eventRecordOf(seqNr = SeqNr.unsafe(2), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(3))).event,
          ),
        )
      } yield {}

      val expected = State(
        actions = List(
          Action.UpdateSeqNr(key, metaSegment, PartitionOffset(Partition.min, Offset.unsafe(3)), timestamp1, SeqNr.unsafe(2)),
          Action.InsertRecords(key, SegmentNr.journal(SeqNr.unsafe(1), segmentSize), 1),
          Action.UpdateSeqNr(key, metaSegment, PartitionOffset(Partition.min, Offset.unsafe(2)), timestamp1, SeqNr.unsafe(1)),
          Action.InsertRecords(key, SegmentNr.journal(SeqNr.unsafe(2), segmentSize), 1),
          Action.InsertMetaJournal(
            key,
            metaSegment,
            created = timestamp0,
            updated = timestamp0,
            journalHead = JournalHead(
              partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(1)),
              segmentSize     = segmentSize,
              seqNr           = SeqNr.unsafe(2),
              deleteTo        = SeqNr.unsafe(2).toDeleteTo.some,
            ),
            origin = none,
          ),
        ),
        metaJournal = Map(
          (
            (topic0, metaSegment),
            Map(
              (
                id,
                MetaJournalEntry(
                  journalHead = JournalHead(
                    partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(3)),
                    segmentSize     = segmentSize,
                    seqNr           = SeqNr.unsafe(2),
                    deleteTo        = SeqNr.unsafe(2).toDeleteTo.some,
                    expiry          = none,
                    recordId        = none,
                  ),
                  created = timestamp0,
                  updated = timestamp1,
                  origin  = none,
                ),
              ),
            ),
          ),
        ),
        journal = Map( // <=== from perspective of `metajournal` table, this is expected to be `empty` - leak
          (
            (key, SegmentNr.journal(SeqNr.unsafe(1), segmentSize)),
            Map(
              (
                (SeqNr.unsafe(1), timestamp0),
                eventRecordOf(seqNr = SeqNr.unsafe(1), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(2)))
                  .copy(metaRecordId = none),
              ),
              (
                (SeqNr.unsafe(2), timestamp0),
                eventRecordOf(seqNr = SeqNr.unsafe(2), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(3)))
                  .copy(metaRecordId = none),
              ),
            ),
          ),
        ),
      )
      val result = stateT.run(State.empty)
      result shouldEqual (expected, ()).pure[Try]
    }
```

mr-git: honestly, I am not sure what is better:
* leaking events in original code
* adding and removing them in code as it is now in this PR

the consistency is better now, though load on Cassandra is not optimal

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,139 +24,83 @@ private[journal] object Batch {
...
+        this.appends.flatMap { appends =>
+          val deleteTo = this.delete.map(_.to.value)
+          val records  = appends.records
+          val actions =
+            if (deleteTo.contains(records.head.action.range.to) && records.tail.nonEmpty) appends.records.tail
+            else appends.records.toList
+          NonEmptyList.fromList(actions) match {
+            case Some(actions) => appends.copy(records = actions).some
+            case None          => appends.some // cannot happen
           }
```
rtar, line 91:
```suggestion
          val actions =
            if (deleteTo.contains(records.head.action.range.to)) NonEmptyList.fromList(records.tail).getOrElse(records)
            else records
          appends.copy(records = actions).some
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -14,139 +24,83 @@ private[journal] object Batch {
...
-              }
-            }
+  private final case class State(
+    purge: Option[Purge]     = None,
+    appends: Option[Appends] = None,
+    delete: Option[Delete]   = None,
+  ) {
+
+    def batches: List[Batch] = {
+      // we can drop first `append`, if `deleteTo` will discard it AND there is at least one more `append`
```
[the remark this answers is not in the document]
mr-git: I added extra comment in code:
```scala
// we have to preserve one `append` to store latest `seqNr` and populate `expireAfter`
```

the alternative was to introduce dedicated new "action" to update only `seqNr` and `expireAfter` fields in `metajournal`, but that was too invasive (in #676 there was an attempt to do so, but it didn't handle `expireAfter` case)

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: hm.. `ReplicatedCassandra` doesn't check if `seqNr` of new event is before `deleteTo` for this journal... :garold:

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: moved inside `State` - did it in two steps - at first created 3 extra functions (`handle*`), but then the code just asked for moving also the `switch` statement inside...

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: following test-case for actions `Delete(12), Append(11), Append(12)` on head with `deleteTo: 10, seqNr:10` "leaks" journal events: `11`, `12` and fails to update `deleteTo` in head - it stays at initial value `10` - code in `ReplicatedCassandra` doesn't allow "delete in future" and clamps attempt down to `seqNr`.

<details>
<summary></summary>

```scala
    test(s"batched `[delete(12), append(11), append(12)]` on active heat must update tables correctly, $suffix") {
      val id          = "id"
      val key         = Key(id = id, topic = topic0)
      val metaSegment = segmentOfId.metaJournal(key)

      val initial = State(
        metaJournal = Map(
          (
            (topic0, metaSegment),
            Map(
              (
                id,
                MetaJournalEntry(
                  journalHead = JournalHead(
                    partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(11)),
                    segmentSize     = segmentSize,
                    seqNr           = SeqNr.unsafe(10L),
                    deleteTo        = SeqNr.unsafe(10L).toDeleteTo.some,
                    expiry          = none,
                    recordId        = none,
                  ),
                  created = timestamp0,
                  updated = timestamp0,
                  origin  = none,
                ),
              ),
            ),
          ),
        ),
      )

      val stateT = for {
        _ <- journal.delete(
          key       = key,
          partition = Partition.min,
          offset    = Offset.unsafe(12),
          timestamp = timestamp1,
          deleteTo  = SeqNr.unsafe(12).toDeleteTo,
          origin    = none,
        )
        _ <- journal.append(
          key         = key,
          partition   = Partition.min,
          offset      = Offset.unsafe(13),
          timestamp   = timestamp1,
          expireAfter = none,
          events = Nel.of(
            eventRecordOf(seqNr = SeqNr.unsafe(11), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(13))).event,
          ),
        )
        _ <- journal.append(
          key         = key,
          partition   = Partition.min,
          offset      = Offset.unsafe(14),
          timestamp   = timestamp1,
          expireAfter = none,
          events = Nel.of(
            eventRecordOf(seqNr = SeqNr.unsafe(12), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(14))).event,
          ),
        )
      } yield {}

      val expected = State(
        actions = List(
          Action.UpdateSeqNr(key, metaSegment, PartitionOffset(Partition.min, Offset.unsafe(14)), timestamp1, SeqNr.unsafe(12)),
          Action.InsertRecords(key, SegmentNr.journal(SeqNr.unsafe(12), segmentSize), 1),
          Action.UpdateSeqNr(key, metaSegment, PartitionOffset(Partition.min, Offset.unsafe(13)), timestamp1, SeqNr.unsafe(11)),
          Action.InsertRecords(key, SegmentNr.journal(SeqNr.unsafe(11), segmentSize), 1),
          Action.UpdatePartitionOffset(key, metaSegment, PartitionOffset(Partition.min, Offset.unsafe(12)), timestamp1),
        ),
        metaJournal = Map(
          (
            (topic0, metaSegment),
            Map(
              (
                id,
                MetaJournalEntry(
                  journalHead = JournalHead(
                    partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(14)),
                    segmentSize     = segmentSize,
                    seqNr           = SeqNr.unsafe(12),
                    deleteTo        = SeqNr.unsafe(10).toDeleteTo.some,
                    expiry          = none,
                    recordId        = none,
                  ),
                  created = timestamp0,
                  updated = timestamp1,
                  origin  = none,
                ),
              ),
            ),
          ),
        ),
        journal = Map(
          (
            (key, SegmentNr.journal(SeqNr.unsafe(12), segmentSize)),
            Map(
              (
                (SeqNr.unsafe(11), timestamp0),
                eventRecordOf(seqNr = SeqNr.unsafe(11), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(13)))
                  .copy(metaRecordId = none),
              ),
              (
                (SeqNr.unsafe(12), timestamp0),
                eventRecordOf(seqNr = SeqNr.unsafe(12), partitionOffset = PartitionOffset(Partition.min, Offset.unsafe(14)))
                  .copy(metaRecordId = none),
              ),
            ),
          ),
        ),
      )
      val result = stateT.run(initial)
      result shouldEqual (expected, ()).pure[Try]
    }
```
</details>

mr-git: the existing code with batch `Delete(12), Append(11),  Append(12)` would:
* for existing journal:
** leave head with `deleteTo: 10, seqNr: 12`
** "leak" 2 events
* for non-existing journal:
** create journal with `deleteTo: 12, seqNr: 12`
** "leak" 2 events without updating `metajournal` table

new code would optimize `Delete(12), Append(11),  Append(12)` to `Append(11),  Append(12), Delete(12)` and:
* for both existing journal and non-existing one:
** update head with `deleteTo: 12, seqNr: 12`
** add and remove event `12`

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -13,140 +23,96 @@ private[journal] sealed abstract class Batch extends Product {
...
+    appends: Option[Appends] = None,
+    delete: Option[Delete]   = None,
+  ) {
 
-    def dropDeleted(appends: Appends, delete: Action.Delete): Option[Appends] = {
-      // we cannot drop individual events in record as we do not parse the payload here (`records.head.action.payload: ByteVector`)
-      val validRecords = appends.records.collect { case record if record.action.range.to >= delete.to.value => record }
-      NonEmptyList.fromList(validRecords).map { Appends(appends.offset, _) }
-    }
+    def handle: ActionRecord[Action] => State = {
```
rtar, line 35: I wonder why this:
```scala
def handle: ActionRecord[Action] => State = {
```

Not this:
```
def handle(record: ActionRecord[Action]): State = record match {
```

I have no preference myself, just curious.

rtar, line 35: As this `handle` method relies on the records being handled in reverse order, it might be worth to name it accordingly or just mention it in the comments.

rtar (approved):

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: only to not have `record match {}` wrapper :) 

probably I could create a `PartialFunction` with only `case`-s inside, but the signature would not be so nice

benchmark/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchBenchmark.scala
```diff
@@ -0,0 +1,364 @@
...
+@Measurement(iterations = 5, time = 10, timeUnit = TimeUnit.SECONDS)
+class BatchBenchmark {
+
+  @Benchmark
+  def original(blackhole: Blackhole): Unit = {
+    val of = Batch_4_1_0.of
+    callAll[Batch_4_1_0](blackhole, of)
+  }
+
+//  @Benchmark
```
migesok, line 28: Please either remove commented out code or leave a code comment above it on what's its purpose and why it should stay in the source code. Just to keep the codebase tidy.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -5,6 +5,16 @@ import cats.syntax.all.*
...
 
+/**
+ * Receives list of records from [[ReplicateRecords]], groups and optimizes similar or sequential actions, like:
+ *  - two or more `append` or `delete` actions are merged into one
+ *  - optimizes list of records to minimize load on Cassandra, like:
+ *    - aggregate effective actions by processing them from the youngest record down to oldest
+ *    - ignore/drop all actions, which are before last `Purge`
+ *    - ignore all `Mark` actions
+ *    - if `append`(s) are followed by `delete` all `append`(s), except last, are dropped
+ *    - at the end, apply all aggregated batches following order: `purge`, `append`s, `delete`
```
migesok, line 16: please describe in the scaladoc the assumptions you make regarding input data with which this algorithm valid (ie it won't work correctly if deletes are made into future)

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -13,140 +30,103 @@ private[journal] sealed abstract class Batch extends Product {
...
-      // we cannot drop individual events in record as we do not parse the payload here (`records.head.action.payload: ByteVector`)
-      val validRecords = appends.records.collect { case record if record.action.range.to >= delete.to.value => record }
-      NonEmptyList.fromList(validRecords).map { Appends(appends.offset, _) }
-    }
+  private final case class State(
+    private val purge: Option[Purge]     = None,
+    private val appends: Option[Appends] = None,
+    private val delete: Option[Delete]   = None,
+  ) {
+    // Expects records to be provided in revered order, e.g. youngest first
```
rtar, line 48: `revered` -> `reversed`

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -13,140 +30,103 @@ private[journal] sealed abstract class Batch extends Product {
...
-        }
+
+      // if `delete` was not last action, adjust `delete`'s batch offset to update `metajournal` correctly
+      val delete = appends match {
+        case Some(appends) => this.delete.map(delete => delete.copy(offset = delete.offset max appends.offset))
+        case None          => this.delete
       }
+
+      // apply action batches in order: `Purge`, `Append`s and `Delete`
+      List(purge, appends, delete).flatten
```
mr-git, line 128: I run extra validation on our load-testing environment with code like:
```scala
      (delete, appends) match {
        case (Some(delete), Some(appends)) if delete.to.value > appends.records.last.action.range.to =>
          log.warn("unexpected situation: `delete.to > appends.seqNr`; falling back to previous batching version")
          // use version of `Batch` from current `master` branch
        case _ =>
          List(purge, appends, delete).flatten
      }
```

over 20+ hours there were no cases, when previous version was used - all our internal clients issue `delete` action with explicitly set `seqNr` which is aligned with aggregate's state (doesn't attempt to "delete in future")

rtar: I just wanted to double check that `log.warn` does not return `IO[Unit]` in this case :thinking:

mr-git: no, it doesn't - there were more lines instead of ```// use version of `Batch` from current `master` branch``` - I just didn't want to post them all here 😄 

I just wanted to be sure, that we do not have active clients, which use `delete(max)`, though it is not expected to be used in "normal" case

it might be, that we have some weird "reset logic" where `delete(max)` makes sense and is used rarely - and this PR would break such journal, if `delete(max)` and next `append` would happen to land in same `poll`

using `delete(max)` makes sense only when journal is broken and somebody manually produces such action with hope that blank state would allow to blindly "reset" aggregate, while explicit `purge` could make more sense

mr-git: I am wondering, if there is a need for explicit handling in case of `delete(max)`?

Technically, if journal has `seqNr: 7`, then `delete(13)` would also act as `delete(max)` - thus special handling for `delete(max)` would be incomplete and `Batch` has no visibility over current state of journal and has no means to check if `delete(x)` would overshoot current `seqNr`.

In addition, `delete(x)` is legit way to [re]start journal from `seqNr: x` - thus one has to be very sure that journal exists in Cassandra before it can use `delete(max)` as otherwise one would create empty journal with no way to publish even one one additional event - `SeqNr.max + 1` would generate exception due to overflow.

It should be safe to merge this PR:
* there is no good reason to overshoot with `Delete` action in regular flow and if business logic tries to overshoot with `Delete`, then it means that it lacks consistency
* if one publishes `Delete` action manually, then one has breached single-writer principle and must be fully aware of internals of `kafka-journal` - thus should be able to act in informed way or be ready to do manual cleanup afterwards

journal/src/main/scala/com/evolutiongaming/kafka/journal/Journal.scala
```diff
@@ -32,7 +32,7 @@ trait Journal[F[_]] {
   /**
    * Deletes events up to provided SeqNr
    */
-  def delete(to: DeleteTo = DeleteTo.max): F[Option[PartitionOffset]]
+  def delete(to: DeleteTo): F[Option[PartitionOffset]]
```
dfakhritdinov, line 35: Removing default value is binary incompatible change. The trait is public thus someone outside of KJ uses it, be careful.

IMO its better to keep `to: DeleteTo = DeleteTo.max` and change on next major version bump.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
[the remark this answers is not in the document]
dfakhritdinov: Sorry but I don't get the answer on @rtar question: why only _first_ event dropped but not _all (until `deleteTo`)_ except one to carry over `seqNr` and `expireAfter`?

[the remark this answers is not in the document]
mr-git: because we drop all of them before during batching, but we keep the `Append` where `range.to == deleteTo.seqNr` - we might add more `append`s after it and if we do add more, then we can drop the first one - all other ones will have `seqNr > deleteTo.seqNr` and we have to store all other `append`s in `journal` table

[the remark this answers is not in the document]
mr-git: example:
```
append(5)
delete(6)
append(6)
append(7)
```
leads to batch: `appends(6, 7); delete(6)` - we have dropped `append(5)` and we can still drop `append(6)`.

in case if there would not be `append(7)`, we would have to keep `append(6)`

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -5,6 +5,23 @@ import cats.syntax.all.*
...
 
+/**
+ * Receives list of records from [[ReplicateRecords]], groups and optimizes similar or sequential actions, like:
+ *  - two or more `append` or `delete` actions are merged into one
+ *  - optimizes list of records to minimize load on Cassandra, like:
+ *    - aggregate effective actions by processing them from the youngest record down to oldest
+ *    - ignore/drop all actions, which are before last `Purge`
+ *    - ignore all `Mark` actions
+ *    - if `append`(s) are followed by `delete` all `append`(s), except last, are dropped
+ *    - at the end, apply all aggregated batches following order: `purge`, `append`s, `delete`
```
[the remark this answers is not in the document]
mr-git: yes - this is the minimal set of different batches with maximum compaction - I added more docs with explanation of reasoning

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -13,140 +30,103 @@ private[journal] sealed abstract class Batch extends Product {
...
-            }
-
-          case Nil =>
-            record.action match {
-              case a: Action.Append => appendsOf(NonEmptyList.of(actionRecord(a))) :: Nil
-              case _: Action.Mark   => Nil
-              case a: Action.Delete => deleteOf(a.to, a.origin, a.version) :: Nil
-              case a: Action.Purge  => purgeOf(a.origin, a.version) :: Nil
-            }
+    def batches: List[Batch] = {
```
dfakhritdinov, line 111: The idea of creating at max three batches out of any number of events leads to splitting the logic on two parts: one for combining events into "batch"es (actually it should not be called batch anymore because its not just collapsed operations but rather instructions for underline logic, not directly connected with original events. As an example check `Batch.Delete` with offset coming from append) and another for hacking some values to fit preserved order (purge, appends, delete) into logic of persisting events in `ReplicatedCassandra`.

I think its not a way we should go

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
```diff
@@ -13,140 +30,103 @@ private[journal] sealed abstract class Batch extends Product {
...
-                case _: Batch.Appends => none
-                case b: Batch.Purge   => b.origin
-              }
-            }
-          }
-        }
+    private def handleDelete(delete: Action.Delete, partitionOffset: PartitionOffset): State = {
+      val delete_ = this.delete match {
+        case Some(younger) =>
+          // take `origin` and `version` from "older" entity, if it has them
```
dfakhritdinov, line 79: Please elaborate _why_ `origin` and `version` are taken from "older" entity

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
dfakhritdinov, line 132: I totally agree that purge must be executed first, but I have strong doubts about doing append before delete because:

1. we're not interested in appends before delete (excluding the case of last append for propagating `seqNr` and `expiration`)
2. all (not dropped) appends actually coming _after_ delete

I propose to change the order of batches to: purge, delete, append* where delete batch will contains not only delete data but optionally last append before the delete. This will remove need in any recalculations of delete.offset, simplifies merging deletes and makes order of batches equal to original events order

dfakhritdinov (changes requested):

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: I was under impression that we do not want to change the "contract" beyond `Batch` and its users, but I can rework code according to your proposal, though then I have to change the "shape" of the `Delete`

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/Batch.scala
mr-git: from what I understood - we want to save the `origin` and `version` of *first* action, which creates the new journal (strange, that original test cases do not check if `append`s keep it correctly - maybe in initial version `delete` didn't have `origin` and `version`?)

mr-git: probably, the `Option` part is for backwards compatibility with older client code - eventually we should remove the optionality

mr-git:
closed in favor of less aggressive #695

PR #695, opened 2024-11-04
Batching with preserved offset ordering

dfakhritdinov:
Based on https://github.com/evolution-gaming/kafka-journal/pull/687, so kudos going to @mr-git 

Main difference, compared with original PR, is in Kafka offset ordering that is preserved in resulted batches.


replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
```diff
@@ -41,164 +41,170 @@ class BatchSpec extends AnyFunSuite with Matchers {
...
         Nel.of(
           append(offset = 1, seqNr = 1),
-          delete(offset = 2, seqNr = 1, origin = "origin1"),
+          delete(offset = 2, to    = 1, origin = "origin1"),
           append(offset = 3, seqNr = 2),
-          delete(offset = 4, seqNr = 2, origin = "origin2"),
+          delete(offset = 4, to    = 2, origin = "origin2"),
         ),
-        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin1")),
+        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, to = 2, origin = "origin2")),
```
mr-git, line 72: here we lose the `origin`

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
```diff
@@ -41,164 +41,170 @@ class BatchSpec extends AnyFunSuite with Matchers {
...
         Nel.of(
           append(offset = 1, seqNr = 1),
-          delete(offset = 2, seqNr = 1, origin = "origin"),
+          delete(offset = 2, to    = 1, origin = "origin"),
           append(offset = 3, seqNr = 2),
-          delete(offset = 4, seqNr = 2),
+          delete(offset = 4, to    = 2),
         ),
-        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin")),
+        List(appends(3, append(offset = 3, seqNr = 2)), deletes(offset = 4, to = 2)),
```
mr-git, line 81: here we lose the `origin`

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
```diff
@@ -41,164 +41,170 @@ class BatchSpec extends AnyFunSuite with Matchers {
...
         Nel.of(
           append(offset = 1, seqNr = 1),
           append(offset = 2, seqNr = 2),
-          delete(offset = 3, seqNr = 1, origin = "origin1"),
-          delete(offset = 4, seqNr = 2, origin = "origin2"),
+          delete(offset = 3, to    = 1, origin = "origin1"),
+          delete(offset = 4, to    = 2, origin = "origin2"),
         ),
-        List(appends(2, append(offset = 2, seqNr = 2)), deletes(offset = 4, seqNr = 2, origin = "origin1")),
+        List(appends(2, append(offset = 2, seqNr = 2)), deletes(offset = 4, to = 2, origin = "origin2")),
```
mr-git, line 90: here we lose the `origin`

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
```diff
@@ -41,164 +41,170 @@ class BatchSpec extends AnyFunSuite with Matchers {
...
           mark(offset   = 2),
-          delete(offset = 3, seqNr = 1, origin = "origin"),
+          delete(offset = 3, to = 1, origin = "origin"),
           mark(offset   = 4),
-          delete(offset = 5, seqNr = 2),
+          delete(offset = 5, to = 2),
           mark(offset   = 6),
         ),
-        List(deletes(offset = 6, seqNr = 2, origin = "origin")),
+        List(deletes(offset = 5, to = 2)),
```
mr-git, line 117: here we lose the `origin`

mr-git: for all above case with "here we lose the `origin`", very possibly we lose `version` too.

basically it means that `origin` and `version`, probably, will get set by one of following `append` actions, instead of [possibly] first-journal-creating `delete` action

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/BatchSpec.scala
dfakhritdinov: I did it on purpose, however I'm not confident was it right decision. Could you pls review the PR ignoring `origin`? Later, if the PR will be accepted - I will back to `origin`

mr-git: I am starting to lean on idea that we should just update `origin` and `version` in `metajournal` on any action, if current version is `None` - the fact that client was able to restore state from some older journal means that it is compatible with current version of kafka-journal - thus we can update.

Eventually all actively used aggregates will have them both set, though we still might have some old and stale journals without `origin` and `version` values

Honestly, I do not see, yet, good reason to aim for very first action, but I might not see some edge case, where it is important.

I think that:
* `origin` was introduced to help with manual untangling in case of split-brain
* `version` was introduced to provide support for possible future journal upgrades or compatibility checks

mr-git (approved):
I do not see, how `origin` and `version` from "second" `delete` action could affect us negatively

PR #697, opened 2024-11-06
make `version` non-optional

mr-git:


journal/src/main/scala/com/evolutiongaming/kafka/journal/ActionHeader.scala
```diff
@@ -37,15 +37,57 @@ object ActionHeader {
...
             metadata    <- metadata
           } yield {
-            Append(range, origin, version, payloadType, metadata)
+            Append(range, origin, version.getOrElse(Version.obsolete), payloadType, metadata)
+          }
+        }
+      }
+      OFormat(reads, format)
+    }
+    val deleteFormat = {
```
dfakhritdinov, line 46: Could you somehow avoid having hand-crafter formatters? I have doubts that I wanna see them in a long term

mr-git: yes, I do not like them too - we could provide default value and use formatter whish supports default values

dfakhritdinov (approved):

journal/src/main/scala/com/evolutiongaming/kafka/journal/ActionHeader.scala
```diff
@@ -96,25 +72,25 @@ object ActionHeader {
   final case class Append(
     range: SeqRange,
     origin: Option[Origin],
-    version: Option[Version],
+    version: Version = Version.obsolete,
```
mr-git, line 75: Technically, we do not need here and in 3 other places the default value, right?

All clients with versions since 0.0.152 have to provide these in payload over Kafka. Oldest client we have is 0.0.177, IIRC, thus all of them must provide `version` field, right?

I checked the history for those `journal/src/test/resources/com/evolutiongaming/kafka/journal` and related unit-tests - they are more than 6 years old, they were introduced in version 0.0.10 - I am leaning on upgrading them to state/format, which is expected at 0.0.177 - then we will require these fallbacks

Other option would be to introduce protocol version, but that could be overkill at this moment

dfakhritdinov:
This PR contains braking changes in public API (including some `case class`es). I see no nice & simple option to overcome this by making desired change backward compatible and propose postponing the PR til next major version bump

stasimus:
@migesok do you want to progress with new version?

PR #738, opened 2025-03-27
Fix TopicReplicator log level

migesok:
Reduce the severity for log statements issued on each consumer poll from INFO to TRACE


migesok:
> why `trace` and not `debug` severity?

I will make another MR where offset commits will be logged to `debug` so there will be 3 levels there:
- `trace` for per-poll activity
- `debug` for periodic low rate activity like offset commits
- `info` for rebalance events

PR #739, opened 2025-03-31
New improved Kafka offset commit logic for replicator

migesok:
The next release version should have a minor increment - 4.2.0

- TopicCommit.apply renamed to TopicCommit.sync, TopicCommit.apply deprecated
- Added TopicCommit.asyncPeriodic implementation
- TopicCommit.delayed deprecated in favor of TopicCommit.asyncPeriodic
- TopicCommit.asyncPeriodic is used by default in the replicator
- updated skafka for an async commit method fix

Main features of asyncPeriodic:
- periodic commits are done asynchronously without blocking the main poll loop, improving latency
- it properly handles rebalance, not loosing progress for revoked partitions
- same logic commits offsets on app shutdown because Java consumer invokes revoke partition callbacks on closure
- gracefully handles RebalanceInProgressException

Additionally:
- made TopicCommit journal package private - it is used only in journal package private APIs, and it is not supposed to be used outside the internal journal code
- ToTry[StateT] impl in ConsumeTopicTest - it didn't work on flatMap'ed RebalanceCallback's

### DIFF
--- a/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
+++ b/replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
@@ -1,44 +1,133 @@
 package com.evolutiongaming.kafka.journal.replicator
 
 import cats.Applicative
-import cats.data.NonEmptyMap as Nem
-import cats.effect.kernel.Concurrent
-import cats.effect.{Clock, Ref}
+import cats.data.{NonEmptyMap as Nem, NonEmptySet}
+import cats.effect.*
 import cats.syntax.all.*
 import com.evolutiongaming.catshelper.ClockHelper.*
 import com.evolutiongaming.catshelper.DataHelper.*
+import com.evolutiongaming.catshelper.Log
 import com.evolutiongaming.kafka.journal.KafkaConsumer
+import com.evolutiongaming.kafka.journal.replicator.commit.AsyncPeriodicTopicCommit
 import com.evolutiongaming.kafka.journal.util.TemporalHelper.*
 import com.evolutiongaming.skafka.*
+import com.evolutiongaming.skafka.consumer.RebalanceCallback
 
 import java.time.Instant
+import scala.annotation.nowarn
 import scala.collection.immutable.SortedMap
 import scala.concurrent.duration.*
 
-trait TopicCommit[F[_]] {
+/**
+ * Kafka partition offset commit logic for replicator internal topic processing API
+ * ([[TopicConsumer]], [[ConsumeTopic]]).
+ *
+ * @tparam F effect type
+ */
+private[journal] trait TopicCommit[F[_]] {
+// journal-private because the places where it is used are journal-private - TopicConsumer, ConsumeTopic
 
+  /**
+   * Mark offsets for commit.
+   * 
+   * The method should be called after each processing step, i.e., after each consumer poll result processing.
+   * 
+   * Please note that offsets passed here should follow the same logic as for Java consumer commit methods:
+   * if offset N is known to be processed, offset N+1 should be committed - offset from which you want to start your
+   * processing next time.
+   * 
+   * Depending on the underlying implementation, this method might execute commit synchronously with the call or
+   * schedule it for later.
+   * 
+   * @param offsets partition offsets to commit
+   */
   def apply(offsets: Nem[Partition, Offset]): F[Unit]
+
+  /**
+   * Notify topic commit logic about new partitions assigned during rebalance - called in
+   * [[com.evolutiongaming.skafka.consumer.RebalanceListener1.onPartitionsAssigned]].
+   * 
+   * @param partitions assigned partitions
+   *                   
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsAssigned(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.empty
+
+  /**
+   * Notify topic commit logic about partitions revoked during rebalance - called in
+   * [[com.evolutiongaming.skafka.consumer.RebalanceListener1.onPartitionsRevoked]].
+   * 
+   * Could be used to commit offsets for revoked partitions using
+   * [[RebalanceCallback.commit]] in order not to lose progress during rebalance.
+   *
+   * @param partitions revoked partitions
+   *
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsRevoked(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.empty
+
+  /**
+   * Notify topic commit logic about partitions lost - called in
+   * [[com.evolutiongaming.skafka.consumer.RebalanceListener1.onPartitionsLost]].
+   *
+   * When partitions are lost as opposed to being revoked:
+   * [[org.apache.kafka.clients.consumer.ConsumerRebalanceListener#onPartitionsLost(java.util.Collection)]]
+   * 
+   * Usually nothing could be done in this case apart from removing the lost partitions from internal state, if there
+   * is any.
+   * As another consumer might already handle the lost partitions by this point, it is highly recommended not to
+   * try to commit offsets for those.
+   *
+   * @param partitions lost partitions
+   *
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsLost(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.empty
 }
 
-object TopicCommit {
+private[journal] object TopicCommit {
 
-  def empty[F[_]: Applicative]: TopicCommit[F] = (_: Nem[Partition, Offset]) => ().pure[F]
+  def empty[F[_]: Applicative]: TopicCommit[F] = new EmptyTopicCommit[F]
 
+  @deprecated("renamed to `sync`, use it instead", since = "4.2.0")
   def apply[F[_]](
     topic: Topic,
     metadata: String,
-    consumer: KafkaConsumer[F, _, _],
-  ): TopicCommit[F] = { (offsets: Nem[Partition, Offset]) =>
-    {
-      val offsets1 = offsets.mapKV { (partition, offset) =>
-        val offset1    = OffsetAndMetadata(offset, metadata)
-        val partition1 = TopicPartition(topic, partition)
-        (partition1, offset1)
-      }
-      consumer.commit(offsets1)
-    }
+    consumer: KafkaConsumer[F, ?, ?],
+  ): TopicCommit[F] = {
+    sync[F](
+      topic          = topic,
+      commitMetadata = metadata,
+      consumer       = consumer,
+    )
   }
 
+  /**
+   * [[TopicCommit]] which performs synchronous blocking commit on each offset mark call.
+   * 
+   * @param topic processed topic name
+   * @param commitMetadata metadata to pass to each consumer commit call
+   * @param consumer [[KafkaConsumer]] to call commit on
+   * @tparam F effect type
+   */
+  def sync[F[_]](
+    topic: Topic,
+    commitMetadata: String,
+    consumer: KafkaConsumer[F, ?, ?],
+  ): TopicCommit[F] = new SyncTopicCommit[F](
+    topic          = topic,
+    commitMetadata = commitMetadata,
+    consumer       = consumer,
+  )
+
+  @deprecated(
+    "use `asyncPeriodic` instead - the delayed impl blocks the poll-loop on commit and " +
+      "looses progress on rebalance and shutdown",
+    since = "4.2.0",
+  )
   def delayed[F[_]: Concurrent: Clock](
     delay: FiniteDuration,
     commit: TopicCommit[F],
@@ -51,9 +140,9 @@ object TopicCommit {
       stateRef  <- Ref[F].of(State(timestamp + delay))
     } yield {
       new TopicCommit[F] {
-        def apply(offsets: Nem[Partition, Offset]) = {
+        def apply(offsets: Nem[Partition, Offset]): F[Unit] = {
 
-          def apply(state: State, timestamp: Instant) = {
+          def apply(state: State, timestamp: Instant): F[State] = {
             val offsets1 = state.offsets ++ offsets.toSortedMap
             if (state.until <= timestamp) {
               offsets1
@@ -75,6 +164,61 @@ object TopicCommit {
           } yield {}
         }
       }
+    }
+  }
+
+  /**
+   * [[TopicCommit]] which commits progress periodically using an async commit method, without blocking the main
+   * poll loop.
+   *
+   * Properly handles rebalance events - when partitions are revoked, it performs a synchronous commit for revoked
+   * partitions in order not to lose the progress.
+   * Since the partition revoke callback is also called on consumer shutdown, the same logic saves the progress on
+   * app shutdown.
+   *
+   * @param topic processed topic name
+   * @param commitMetadata metadata to pass to each consumer commit call
+   * @param commitPeriod delay between periodic async commit calls
+   * @param consumer [[KafkaConsumer]] to call commit on
+   * @param log logger instance.
+   *            The logic here doesn't add anything to log statements which can identify the topic or particular
+   *            consumer instance.
+   *            If you have more than one topic-processor in the app, add that information to the logger before passing
+   *            it here.
+   * @tparam F effect type
+   * @return implementation wrapped in Resource - don't forget to release it after use!
+   */
+  def asyncPeriodic[F[_]: Temporal](
+    topic: Topic,
+    commitMetadata: String,
+    commitPeriod: FiniteDuration,
+    consumer: KafkaConsumer[F, ?, ?],
+    log: Log[F],
+  ): Resource[F, TopicCommit[F]] =
+    AsyncPeriodicTopicCommit.make(
+      topic          = topic,
+      commitMetadata = commitMetadata,
+      commitPeriod   = commitPeriod,
+      consumer       = consumer,
+      log            = log,
+    )
+
+  private final class EmptyTopicCommit[F[_]: Applicative] extends TopicCommit[F] {
+    override def apply(offsets: Nem[Partition, Offset]): F[Unit] = ().pure
+  }
+
+  private final class SyncTopicCommit[F[_]](
+    topic: Topic,
+    commitMetadata: String,
+    consumer: KafkaConsumer[F, ?, ?],
+  ) extends TopicCommit[F] {
+    override def apply(offsets: Nem[Partition, Offset]): F[Unit] = {
+      val offsets1 = offsets.mapKV { (partition, offset) =>
+        val offset1    = OffsetAndMetadata(offset, commitMetadata)
+        val partition1 = TopicPartition(topic, partition)
+        (partition1, offset1)
+      }
+      consumer.commit(offsets1)
     }
   }
 }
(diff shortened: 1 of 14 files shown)


replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+  /**
+   * Notify topic commit logic about new partitions assigned during rebalance - called in
+   * [[com.evolutiongaming.skafka.consumer.RebalanceListener1.onPartitionsAssigned]].
+   * 
+   * @param partitions assigned partitions
+   *                   
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsAssigned(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.pure(())
```
mr-git, line 56: maybe to make it more expressive 
```suggestion
    RebalanceCallback.empty
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+   * 
+   * Could be used to commit offsets for revoked partitions using
+   * [[RebalanceCallback.commit]] in order not to lose progress during rebalance.
+   *
+   * @param partitions revoked partitions
+   *
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsRevoked(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.pure(())
```
mr-git, line 70: expressiveness:
```suggestion
    RebalanceCallback.empty
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+   * 
+   * Usually nothing could be done in this case apart from removing the lost partitions from internal state, if there
+   * is any.
+   *
+   * @param partitions lost partitions
+   *
+   * @return [[RebalanceCallback]]
+   */
+  def onPartitionsLost(@nowarn("cat=unused") partitions: NonEmptySet[Partition]): RebalanceCallback[F, Unit] =
+    RebalanceCallback.pure(())
```
mr-git, line 87: again:
```suggestion
    RebalanceCallback.empty
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+
+  /**
+   * Notify topic commit logic about partitions lost - called in
+   * [[com.evolutiongaming.skafka.consumer.RebalanceListener1.onPartitionsLost]].
+   *
+   * When partitions are lost as opposed to being revoked:
+   * [[org.apache.kafka.clients.consumer.ConsumerRebalanceListener#onPartitionsLost(java.util.Collection)]]
+   * 
+   * Usually nothing could be done in this case apart from removing the lost partitions from internal state, if there
+   * is any.
```
mr-git, line 80: shall we add more letter, like:
```suggestion
   * Usually nothing could be done in this case apart from removing the lost partitions from internal state, if there
   * is any. As lost partitions usually are already handled by other consumer, it is recommended to NOT commit offsets concurrently.
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+    RebalanceCallback.pure(())
 }
 
-object TopicCommit {
+private[journal] object TopicCommit {
 
-  def empty[F[_]: Applicative]: TopicCommit[F] = (_: Nem[Partition, Offset]) => ().pure[F]
+  def empty[F[_]: Applicative]: TopicCommit[F] = new EmptyTopicCommit[F]
 
+  @deprecated("renamed to sync", since = "4.2.0")
```
mr-git, line 94: IMHO message should be about "how to deal with it", not "what was done":
```suggestion
  @deprecated("use `sync` instead", since = "4.2.0")
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
...
+    commitMetadata: String,
+    consumer: KafkaConsumer[F, ?, ?],
+  ): TopicCommit[F] = new SyncTopicCommit[F](
+    topic          = topic,
+    commitMetadata = commitMetadata,
+    consumer       = consumer,
+  )
+
+  @deprecated(
+    "use asyncPeriodic instead - the delayed impl blocks the pool-loop on commit and " +
```
mr-git, line 126: shouldn't `pool` be `poll`?
```suggestion
    "use `asyncPeriodic` instead - the `delayed` implementation blocks the poll-loop on commit and " +
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -77,4 +165,59 @@ object TopicCommit {
       }
     }
   }
+
+  /**
+   * [[TopicCommit]] which commits progress periodically using an async commit method, without blocking the main
+   * pool loop.
```
mr-git, line 171:
```suggestion
   * poll loop.
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicReplicator.scala
```diff
@@ -267,9 +266,15 @@ private[journal] object TopicReplicator {
...
       for {
         consumer <- KafkaConsumerOf[F].apply[String, ByteVector](config1)
+        log      <- topicLoggerOf[F](topic).toResource
         metadata  = hostName.fold { Metadata.empty } { _.value }
-        commit    = TopicCommit(topic, metadata, consumer)
-        commit   <- TopicCommit.delayed(5.seconds, commit).toResource
+        commit <- TopicCommit.asyncPeriodic(
+          topic          = topic,
+          commitMetadata = metadata,
+          commitPeriod   = 5.seconds,
```
mr-git, line 274: these `5.seconds` are not documented, right?

shall we update ScalaDoc or expose it as settable argument with default value?

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
```diff
@@ -0,0 +1,204 @@
+package com.evolutiongaming.kafka.journal.replicator.commit
+
+import cats.Applicative
+import cats.data.{NonEmptyMap, NonEmptySet}
+import cats.effect.*
+import cats.effect.implicits.*
```
mr-git, line 6:
```suggestion
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
```diff
@@ -0,0 +1,204 @@
...
+
+    def removeAssignedPartitions(partitions: Iterable[Partition]): State = {
+      copy(assignedPartitions = assignedPartitions.removedAll(partitions))
+    }
+
+    def applyCommittedForAssigned(offsets: Iterable[(Partition, Offset)]): State = {
+      applyForAssigned(offsets)(_.updateCommittedIfNeeded(_))
+    }
+
+    def applyMarkedForAssigned(offsets: Iterable[(Partition, Offset)]): State = {
```
mr-git, line 151: do we need both: "mark" and "assign" in past? maybe:
```suggestion
    def applyMarkForAssigned(offsets: Iterable[(Partition, Offset)]): State = {
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
```diff
@@ -0,0 +1,204 @@
...
+
+      anythingCommitted <- commitIfNotEmpty[RebalanceCallback[F, *]](revokedToCommit, commitF = RebalanceCallback.commit)
+        .recoverWith {
+          case t: Throwable =>
+            log.error(s"offset commit for revoked partitions on rebalance failed: ${t.getMessage}", t).lift.as(false)
+        }
+
+      _ <-
+        if (anythingCommitted) log.debug(s"offset commit for revoked partitions on rebalance success: $revokedToCommit").lift
+        else RebalanceCallback.pure(())
```
mr-git, line 52: maybe:
```suggestion
        else RebalanceCallback.empty
```

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
```diff
@@ -0,0 +1,204 @@
...
+  ) {
+    def addAssignedPartitions(partitions: Iterable[Partition]): State = {
+      copy(
+        assignedPartitions = partitions.foldLeft(assignedPartitions) { (prevAssigned, partition) =>
+          prevAssigned.updatedWith(partition)(prevStateOpt => prevStateOpt.getOrElse(PartitionState()).some)
+        },
+      )
+    }
+
+    def removeAssignedPartitions(partitions: Iterable[Partition]): State = {
```
mr-git, line 143: maybe:
```suggestion
    def removeRevokedPartitions(partitions: Iterable[Partition]): State = {
```

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/DelayedTopicCommitTest.scala
```diff
@@ -8,15 +8,17 @@ import com.evolutiongaming.skafka.{Offset, Partition}
 import org.scalatest.funsuite.AsyncFunSuite
 import org.scalatest.matchers.should.Matchers
 
+import scala.annotation.nowarn
 import scala.concurrent.duration.*
 
-class TopicCommitTest extends AsyncFunSuite with Matchers {
+@nowarn("cat=deprecation")
```
mr-git, line 14: why we need this?

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/SKafkaTestUtils.scala
```diff
@@ -0,0 +1,11 @@
+package com.evolutiongaming.kafka.journal.replicator
+
+import com.evolutiongaming.skafka.{Offset, Partition, TopicPartition}
+
+object SKafkaTestUtils {
+  def offset(n: Int): Offset = Offset.unsafe(n)
```
mr-git, line 6: why only this one has full nam? shouldn't it be like:
```suggestion
  def o(n: Long): Offset = Offset.unsafe(n)
```

BTW, `Offset` takes `Long`, not `Int`

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
```diff
@@ -1,44 +1,132 @@
 package com.evolutiongaming.kafka.journal.replicator
 
 import cats.Applicative
-import cats.data.NonEmptyMap as Nem
+import cats.data.{NonEmptyMap as Nem, NonEmptySet}
+import cats.effect.*
 import cats.effect.kernel.Concurrent
```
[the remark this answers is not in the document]
migesok: will be fixed

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: I was looking for smth like `.unit` but this works as well. Will be fixed

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: Will be fixed

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: `use sync instead` might suggest something changed in the behaviour and I need to pay attention. I wanted to emphasize that it is only a renaming change, the behaviour is the same.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: Will fix `pool`. Why do you think ticks are needed in deprecation messages? Is it how they do it in the Scala standard library? I don't see right away any value here.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: will be fixed

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: I won't compile - `.toResource` is used in the code here

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: The style has to match the other method: `applyCommittedForAssigned`. If I change it to `applyCommitForAssigned` it would be ambiguous - do I commit here? Or just change the state. I'd rather leave it as is.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: fixed

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/DelayedTopicCommitTest.scala
migesok: `delayed` is marked as deprecated, this whole test lights up with warnings if not suppressed.

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/SKafkaTestUtils.scala
migesok: I use only Int's in tests, no point in accepting Long's here.
`o` looks too ambiguous in the test

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: The state doesn't know anything about "revoked" or "lost" partitions - it just has a set of "assigned" partitions. I don't want to pollute it semantically with complications from higher layers here - i.e. lost vs revoked etc.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: There is a link to `org.apache.kafka.clients.consumer.ConsumerRebalanceListener#onPartitionsLost(java.util.Collection)` above, I don't it is a good idea to duplicate information from it here.

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
mr-git: right, `import import cats.effect.syntax.all.*` is the way to go

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
mr-git: hm.. then I've failed to understand [from name], what the function does

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/SKafkaTestUtils.scala
mr-git: then renaming of `p` to `partition` and `tp` to `topicPartition` should be done to keep consistency

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
mr-git: "removing assigned partitions" feels wrong too - then maybe better to have two functions:
* `addPartitions` or `trackPartitions`
* `removePartitions`?

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
mr-git: IMHO having clearer docs is beneficial, even if it duplicates line, which are about 10 navigation jumps away in IDE

replicator/src/test/scala/com/evolutiongaming/kafka/journal/replicator/SKafkaTestUtils.scala
migesok: no, I like it this way

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: it could be `unassign` but current naming looks OK to me.
- `assignedPartitions` in the state
- `addAssignedPartitions` and `removeAssignedPartitions` for manipulating the set

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/commit/AsyncPeriodicTopicCommit.scala
migesok: fixed

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
mr-git: I like to find important information at 1-click distance, not somewhere far and deep away

mr-git (approved):

replicator/src/main/scala/com/evolutiongaming/kafka/journal/replicator/TopicCommit.scala
migesok: added additional info the doc about concurrent commits

PR #743, opened 2025-04-07
Replace kafka-launcher with testcontainers-kafka

migesok:
This will come in handy when we migrate to Scala 3:
- kafka-launcher starts Kafka in the same JVM and classpath as the KJ code, so we are tied to Kafka server's own Scala version support
- testcontainers Kafka starts in Docker, thus we are not tied anymore

Additionally:
- migrated from a Scala wrapper to using testcontainers directly - this way we can use the latest testcontainers version
- now Kafka and Cassandra start in parallel in tests
- test Cassandra and Kafka versions are now set explicitly

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,27 +2,26 @@ import sbt.*
 
 object Dependencies {
 
-  val scalatest                  = "org.scalatest"          %% "scalatest"                      % "3.2.19"
-  val `scala-java8-compat`       = "org.scala-lang.modules" %% "scala-java8-compat"             % "1.0.2"
-  val `kind-projector`           = "org.typelevel"           % "kind-projector"                 % "0.13.3"
-  val `cassandra-driver`         = "com.datastax.cassandra"  % "cassandra-driver-core"          % "3.11.2"
-  val `play-json`                = "com.typesafe.play"      %% "play-json"                      % "2.9.3"
-  val `play-json-jsoniter`       = "com.evolution"          %% "play-json-jsoniter"             % "1.1.1"
-  val `executor-tools`           = "com.evolutiongaming"    %% "executor-tools"                 % "1.0.3"
-  val `akka-serialization`       = "com.evolutiongaming"    %% "akka-serialization"             % "1.1.0"
-  val `kafka-launcher`           = "com.evolutiongaming"    %% "kafka-launcher"                 % "0.2.0"
-  val `testcontainers-cassandra` = "com.dimafeng"           %% "testcontainers-scala-cassandra" % "0.43.0"
-  val hostname                   = "com.evolutiongaming"    %% "hostname"                       % "0.2.0"
-  val scassandra                 = "com.evolutiongaming"    %% "scassandra"                     % "5.3.0"
-  val `cassandra-sync`           = "com.evolutiongaming"    %% "cassandra-sync"                 % "3.0.0"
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/com/evolutiongaming/kafka/journal/IntegrationSuite.scala
```diff
@@ -14,52 +12,50 @@ import com.evolutiongaming.scassandra.CassandraClusterOf
...
 import com.github.dockerjava.api.model.{ExposedPort, HostConfig, PortBinding, Ports}
 import com.typesafe.config.ConfigFactory
+import org.testcontainers.cassandra.CassandraContainer
+import org.testcontainers.containers.GenericContainer
+import org.testcontainers.kafka.KafkaContainer
+import org.testcontainers.utility.DockerImageName
 
 object IntegrationSuite {
+  private val CassandraDockerImage = DockerImageName.parse("cassandra:4.1.8") // latest 4.x
+  private val KafkaDockerImage     = DockerImageName.parse("apache/kafka-native:3.8.1") // 3.9.0 doesn't start here
```
mr-git, line 22: what about 4.0.0? :)

project/Dependencies.scala
```diff
@@ -2,27 +2,26 @@ import sbt.*
...
+  val scassandra           = "com.evolutiongaming"    %% "scassandra"            % "5.3.0"
+  val `cassandra-sync`     = "com.evolutiongaming"    %% "cassandra-sync"        % "3.0.0"
+  val `cats-helper`        = "com.evolutiongaming"    %% "cats-helper"           % "3.11.3"
+  val random               = "com.evolutiongaming"    %% "random"                % "1.0.0"
+  val retry                = "com.evolutiongaming"    %% "retry"                 % "3.0.1"
+  val sstream              = "com.evolutiongaming"    %% "sstream"               % "1.0.1"
+  val skafka               = "com.evolutiongaming"    %% "skafka"                % "17.1.3"
+  val `akka-test-actor`    = "com.evolutiongaming"    %% "akka-test-actor"       % "0.1.0"
+  val scache               = "com.evolution"          %% "scache"                % "5.1.3"
+  val `resource-pool`      = "com.evolution"          %% "resource-pool"         % "1.0.5"
```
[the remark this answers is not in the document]
migesok: I know, I know.

tests/src/test/scala/com/evolutiongaming/kafka/journal/IntegrationSuite.scala
mr-git, line 22: can we make the version constants somehow exposed for Scala Steward's update suggestions? 

not likely, I guess, but we have to be able somehow to not lose them over time

migesok: I didn't try, not sure I want to. We don't use 4.x Kafka anywhere, I don't think it makes sense to test with it.

migesok: probably only by turning those constants into separate Scala library artefacts, with versions matching container versions. if you think it is worth it, please create a separate issue for it. I'm not sure it is worth it though.

mr-git: what's the problem with 3.9.0?

mr-git (approved):

PR #744, opened 2025-04-08
update formatting

mr-git:


.scalafmt.conf
```diff
@@ -1,51 +1,93 @@
-version = 3.9.4
+# Main goals:
+# - nicer commit diffs (trailing commas, no alignment for pattern matching, force new lines)
+# - better interop with default IntelliJ IDEA setup (matching import and modifiers sorting logic)
+# - better developer experience on laptop screens (like 16' MBPs) with IntelliJ IDEA (line wraps)
 
-project.git = true
```
mr-git, line 3: why discard it? 

https://scalameta.org/scalafmt/docs/configuration.html#projectgit?

.scalafmt.conf
```diff
@@ -1,51 +1,93 @@
...
+#     arg2: String,
+#   ) extends MyTrait {
+#
+# without it:
+#   class MyCls(
+#     arg1: String,
+#     arg2: String) extends MyTrait {
+danglingParentheses.exclude = []
 
-docstrings.style = keep
```
mr-git, line 37: this was added to "force-keep" the formatting of some ScalaDocs, where lists are used - without it ScalaDoc reformats list to paragraph

.scalafmt.conf
```diff
@@ -1,51 +1,93 @@
...
-  { code = "<-", owner = "Enumerator.Generator" }
-  { code = "=", owner = "(Defn.Val|Defn.Var|Type|Def|Enumerator.Val|Assign|Term.Param)" } # Defn.Val is mostly for Dependencies.scala,
-  { code = "->", owner = "Term.ApplyInfix" }
-]
 
-newlines.implicitParamListModifierPrefer = before
-newlines.beforeCurlyLambdaParams = multilineWithCaseOnly
+verticalMultiline {
+  atDefnSite = true
+  arityThreshold = 3 # more than 3 args in a list will be turned vertical
```
mr-git, line 48: this is not `more than 3`, but `at least 3` - IMHO we can go with `4` or `5`

mr-git:
will you update `.git-blame-ignore-revs`? I am not sure, when it should be updated, but it seems that post-factum should be OK

.scalafmt.conf
migesok: added it back

.scalafmt.conf
migesok: why? 3 is OK, multi-line should be preferred for commit-diffs anyways

migesok:
```
This formatting is only triggered if the method definition exceeds the maxColumn value in width or if the number of arguments to the method exceeds the [verticalMultiline.arityThreshold](https://scalameta.org/scalafmt/docs/configuration.html#verticalmultilinearitythreshold)
```

notice "exceeds"

migesok: I double checked, the scalafmt doc lies (
It's >= threshold.
I will change to 4.

mr-git: I noticed that logging code, like `warn(..., ..., ...)` got split to multi-line

overall, I would not mind to have it at 1 :)

mr-git:
APPROVED!

(I cannot press the button as I created the PR from @migesok's branch)

cassandra/src/test/scala/com/evolutiongaming/kafka/journal/cassandra/CreateTablesSpec.scala
```diff
@@ -121,25 +120,37 @@ class CreateTablesSpec extends AnyFunSuite {
...
-    def error(msg: => String, mdc: Log.Mdc)                   = ().pure[F]
-    def error(msg: => String, cause: Throwable, mdc: Log.Mdc) = ().pure[F]
+    def trace(msg: => String, mdc: Log.Mdc) = ().pure[F]
+    def debug(msg: => String, mdc: Log.Mdc) = ().pure[F]
+    def warn(msg: => String, mdc: Log.Mdc) = ().pure[F]
+    def warn(
+      msg: => String,
+      cause: Throwable,
+      mdc: Log.Mdc,
+    ) = ().pure[F]
```
mr-git, line 130: unexpected split?

migesok: fixed with the new threshold config

.scalafmt.conf
migesok: I manually reviewed all scaladoc format changes in files and fixed a couple of places where it broke the markup.

PR #752, opened 2025-04-11
Prepare for Scala 3 support - more code changes

migesok:
Remove type arg variance in PartitionCache.Result
as it trips out Scala 3 compiler in pattern-matching. Also having variance for F[_]-kind of type args
doesn't look right.

### DIFF
--- a/journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
+++ b/journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
@@ -138,7 +138,7 @@ private[journal] object PartitionCache {
      * expiring listener (or deferred value), which returns an actual information if it gets into a
      * cache in a timely manner, or [[Result.Now.Timeout]] if the configured timeout expires.
      */
-    final case class Listener(id: String, offset: Offset, deferred: Deferred[F, Either[Throwable, Result.Now]])
+    final case class Listener(id: String, offset: Offset, deferred: Deferred[F, Either[Throwable, Result.Now[F]]])
 
     type ListenerId = Int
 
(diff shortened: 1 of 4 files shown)


journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
```diff
@@ -527,7 +527,7 @@ private[journal] object PartitionCache {
   /**
    * State of the non-replicated journal head comparing to a given offset.
    */
-  sealed trait Result[+F[_]]
```
mr-git, line 530: I wonder, why it was required previously...

was it only to allow `sealed trait Now extends Result[Nothing]`?

journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
```diff
@@ -576,22 +576,22 @@ private[journal] object PartitionCache {
...
+    def empty[F[_]](value: F[Now[F]]): Result[F] = Later.empty(value)
 
     /**
      * [[PartitionCache]] already seen such [[Offset]] in Kafka or Cassandra.
      *
      * In other words, it means that this offset was either already replicated, or had the latest
      * [[HeadInfo]] information inside of [[PartitionCache]] when [[PartitionCache#get]] was called.
      */
-    sealed trait Now extends Result[Nothing]
+    sealed trait Now[F[_]] extends Result[F]
```
mr-git, line 594: why do we need the type-class here?

journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
migesok: yes, exactly

journal/src/main/scala/com/evolutiongaming/kafka/journal/PartitionCache.scala
migesok: Because Result requires it, you can't extend it without a typeclass

mr-git (approved):

PR #757, opened 2025-04-16
Prepare for Scala 3 support - remove scodec upcast usage

migesok:
In the Events type scodec usage didn't compile with Scala3: upcast method didn't work if a type is wrapped in Option. To fix that, the Events codec was refactored to use the discriminated method.

Also:
- refactored serde tests against file examples to make them more robust and readable
- fixed EventsToBytesSpec not verifying exact byte outputs for the current format which allowed format errors to slip through
- deprecated PayloadType.ext because it was used only in test code and didn't seem to belong in the main model code


persistence/src/test/scala/akka/persistence/kafka/journal/EventSerializerSpec.scala
```diff
@@ -40,32 +34,20 @@ class EventSerializerSpec extends AsyncFunSuite with ActorSuite with Matchers {
...
-          case a: Payload.Binary => writeToFile(a.value, name)
-          case _: Payload.Text   => IO.unit
-          case _: Payload.Json   => IO.unit
-        }*/
-        bytes <- ByteVectorOf[IO](getClass, name)
+//        _ <- payload match {
+//          case a: Payload.Binary => IO { dumpEncodedDataToFile(a.value, name) }
+//          case _: Payload.Text => IO.unit
+//          case _: Payload.Json => IO.unit
+//        }
```
mr-git, line 44: do we need these lines?

journal/src/main/scala/com/evolutiongaming/kafka/journal/Event.scala
```diff
@@ -26,29 +25,14 @@ object Event {
 
     implicit val payloadCodec: Codec[Option[Payload]] = {
 
-      val errEmpty = Err("")
+      // in the original codec, all subtypes were prefixed with an int32 length, even the None option
```
mr-git, line 28: what is `the original codec`?

journal/src/test/scala/com/evolutiongaming/kafka/journal/ActionHeaderJsonSpec.scala
```diff
@@ -8,75 +8,82 @@ import play.api.libs.json.*
...
+        test(s"Append current format, origin: $origin, payloadType: $payloadType, metadata: $metadataStr") {
+          val header = appendExample(origin, payloadType, metadata)
+          verifyEncodeDecodeExample(header, s"Append-$originStr-$payloadType-$metadataStr.json")
+        }
+      }
+
+      metadataExamplesLegacyFormat.foreach { case (metadataStr, metadata) =>
+        test(s"Append legacy format, origin: $origin, payloadType: $payloadType, metadata: $metadataStr") {
+          val header = appendExample(origin, payloadType, metadata)
+          verifyDecodeExample(header, s"Append-$originStr-$payloadType-$metadataStr.json")
```
mr-git, line 59: we have only `read` support for legacy format? 

how old is `legacy`? can we drop it? do we know, if `legacy` is even used?

mr-git (approved):

journal/src/test/scala/com/evolutiongaming/kafka/journal/ActionHeaderJsonSpec.scala
migesok: >we have only read support for legacy format?

How could you support writing a legacy format? It's always reading previous format versions and read-writing the current format version.

mr-git: can we drop the support for "legacy"? I suppose that we cannot, as we cannot know, if there is some stale journal, which hasn't been upgraded, right?

migesok: >how old is legacy? can we drop it? do we know, if legacy is even used?

No idea, I kept the same logic of the test as before. I don't think I want to do soul-searching here - most likely old format are with us forever and ever.

journal/src/main/scala/com/evolutiongaming/kafka/journal/Event.scala
mr-git: now I am lost - is the `original codec` one before this PR? if so, is the `new one` fully compatible?

mr-git (approved):

PR #888, opened 2026-05-26
Extend trait Journal with mark operation

nviliunov-evolution-throwaway:
Currently mark actions are produced for synchronization only during actor recoveries. This allows using mark actions for synchronization in other contexts when it is important that all preceding events have been consumed.

### DIFF
--- a/akka/persistence/src/test/scala/com/evolution/kafka/journal/akka/persistence/JournalAdapterSpec.scala
+++ b/akka/persistence/src/test/scala/com/evolution/kafka/journal/akka/persistence/JournalAdapterSpec.scala
@@ -223,6 +223,12 @@ object JournalAdapterSpec {
             (state1, none[PartitionOffset])
           }
         }
+
+        def mark(id: String): StateT[PartitionOffset] = {
+          StateT { state =>
+            (state, partitionOffset)
+          }
+        }
       }
     }
 
--- a/journal/src/main/scala/com/evolution/kafka/journal/Journal.scala
+++ b/journal/src/main/scala/com/evolution/kafka/journal/Journal.scala
@@ -48,6 +48,16 @@ trait Journal[F[_]] {
    * key, consecutive pointer call will return `None`. Mostly used by [[PurgeExpired]].
    */
   def purge: F[Option[PartitionOffset]]
+
+  /**
+   * Produces a `mark` action with the given correlation id, returning the partition/offset where
+   * the mark was produced.
+   *
+   * Intended for clients that need to observe replication of preceding actions externally, by
+   * subscribing to marks in the replicated topic. The `id` should be unique per use case (callers
+   * are responsible for generating an id with sufficient entropy).
+   */
+  def mark(id: String): F[PartitionOffset]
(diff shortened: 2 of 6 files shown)


journal/src/main/scala/com/evolution/kafka/journal/Journal2.scala
```diff
@@ -0,0 +1,218 @@
...
+import pureconfig.generic.semiauto.*
+
+import scala.concurrent.duration.*
+
+/**
+ * Extension of [[Journal]] that adds [[mark]] for producing a `mark` action with a caller-supplied
+ * correlation id. Kept as a separate trait so adding `mark` to the public `Journal` trait does not
+ * break binary compatibility.
+ */
+trait Journal2[F[_]] extends Journal[F] {
```
mr-git: do we want to be explicit?

```suggestion
trait JournalWithMark[F[_]] extends Journal[F] {
```
OR

do we expect more API exposure?

---
should we plan to merge the `Journal2` (`JournalWithMark`) into `Journal` with major bump?

journal/src/main/scala/com/evolution/kafka/journal/Journal2.scala
```diff
@@ -0,0 +1,218 @@
...
+   * are responsible for generating an id with sufficient entropy).
+   */
+  def mark(id: String): F[PartitionOffset]
+}
+
+object Journal2 {
+
+  def empty[F[_]: Applicative]: Journal2[F] = {
+    class Empty
+    new Empty with Journal2[F] {
```
mr-git: I wonder, if we can make use of `Journal.empty`? can we use `fromJournal` here too?

(I dislike duplicated code)

journal/src/main/scala/com/evolution/kafka/journal/Journal2.scala
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  final case class CallTimeThresholds(
+    append: FiniteDuration = 500.millis,
+    read: FiniteDuration = 5.seconds,
+    pointer: FiniteDuration = 1.second,
+    delete: FiniteDuration = 1.second,
+    purge: FiniteDuration = 1.second,
+    mark: FiniteDuration = 1.second,
+  ) {
```
mr-git: honestly, it is a bit ugly, a lot of duplication

mr-git:
frankly, if we simply extend like in https://github.com/evolution-gaming/kafka-journal/pull/888/changes/0a2243815bd2142d7cee265901447ed8e4b3a50b and make a major (minor?) version bump, the overall code looks much more consistent and cleaner

unless we plan to maintain and extend `Journal2` with more extras, I'd go with simpler code and version bump.

mr-git:
LGTM!

@migesok, what do you think?

migesok:
@nviliunov-evolution-throwaway please fill out the PR description explaining why mark operation should be exposed.  So far it was an internal implementation detail for the Journal.

migesok:
LGTM

PR #889, opened 2026-05-26
Make Journals instance available via actor system extension

nviliunov-evolution-throwaway:
Currently Journal is initialized deep in the persistence plugin and is not available outside of it in application code. The new actor system extension allows retrieving the Journal instance and use it outside the plugin.


mr-git (approved):
LGTM!

@migesok, what do you think?

PR #919, opened 2026-07-18
Migrate build to sbt 2

stasimus:
Updates sbt to 2.0.2, bumps Evolution plugins, drops sbt-coveralls for coveralls GitHub Action. Replaces #910.

PR #920, opened 2026-07-19
Extract shared integration test fixtures

stasimus:
Moves the framework-agnostic test fixtures (IntegrationSuite, CassandraSuite, ReadEventsApp) into a new kafka-journal-tests-shared module, removing the akka/pekko duplication. Based on #919.


mr-git:
Original idea was to [eventually] just discard all `akka` modules in a single commit.

With shared module it will be possible, though one will have to know that those 3 classes can be moved back to their "original" location [now] in only `pekko` module.

Do we really need it?

stasimus:
No, can be closed

Issue #921, opened 2026-07-19
Poison record permanently stalls replicator topic (ConsumeTopic retry has no skip)


stasimus:
## Problem

A single unprocessable ("poison") Kafka record permanently stalls replication of its topic.

In `ConsumeTopic.apply` the poll loop persists records via `topicFlow(records)` and only then marks/commits offsets (`replicator/src/main/scala/com/evolution/kafka/journal/replicator/ConsumeTopic.scala:65-94`). If any record fails conversion, deserialization, or `journal.append`, the whole poll fails and is caught by the retry configured in `ConsumeTopic.of`:

```scala
strategy = Strategy
  .exponential(100.millis)
  .jitter(random)
  .limit(1.minute)      // caps the DELAY only
  .resetAfter(5.minutes)
```
(`ConsumeTopic.scala:26-34`)

There is **no max-attempt limit and no dead-letter/skip path**. On each retry the consumer resource is torn down and re-acquired, re-subscribing and polling from the last committed offset — which returns the same bad record. Because the offset for the bad record is never committed, offsets never advance.

## Failure scenario

One corrupt or newer-schema record on a partition (e.g. an unparseable `journal.action` header) fails on every attempt → the topic consumer loops forever, replication lag grows unbounded, no offsets commit. Requires manual intervention.

## Suggested direction

Bounded retry plus a poison-record quarantine / dead-letter path so a single bad record can be skipped (with logging/metrics) and offsets advance. Related to the HeadCache variant of the same pattern (separate issue).

mr-git:
This is indeed by design - there is no easy way how to quarantine such aggregate - it would require separate storage and some way how to organize the merge back to common storage.

Replicator has no [direct] means to stop the aggregate (the source of events, the "poison"). Sadly, it affects **all** replicated records, but on other hand, when it happens it is very obvious.

There is no easy way to resolve it either, as we cannot replace the message on Kafka, nor we can skip it - we cannot know if next event is valid and applicable.

Only option is to add extra filtering to NOT replicate events from broken aggregate and stop-and-fix the aggregate on client side.

PR #932, opened 2026-07-19
Document why ConsumeTopic never skips records

stasimus:
Adds a Scaladoc to `ConsumeTopic` explaining that stalling on an unprocessable record is intended: the replicator must preserve journal order and completeness, so no record may be skipped. Documents the expected behaviour raised in #921.

### DIFF
--- a/replicator/src/main/scala/com/evolution/kafka/journal/replicator/ConsumeTopic.scala
+++ b/replicator/src/main/scala/com/evolution/kafka/journal/replicator/ConsumeTopic.scala
@@ -14,13 +14,13 @@ import com.evolutiongaming.skafka.consumer.{RebalanceCallback, RebalanceListener
 import scala.concurrent.duration.*
 
 /** Consumes a topic and replicates its records into the eventual store.
-  *
-  * By design there is no skip or dead-letter path. The replicator must preserve the exact order and completeness of
-  * the journal, so no record may ever be dropped. A record that cannot be processed (for example a corrupt or
-  * unparseable one) therefore intentionally stalls replication of its topic — the poll is retried with backoff and
-  * offsets are not advanced past it — until the underlying cause is resolved. This stalling is the expected
-  * behaviour, not a bug: skipping the record would silently diverge the replica from the source journal.
-  */
+ *
+ * By design there is no skip or dead-letter path. The replicator must preserve the exact order and completeness of
+ * the journal, so no record may ever be dropped. A record that cannot be processed (for example a corrupt or
+ * unparseable one) therefore intentionally stalls replication of its topic — the poll is retried with backoff and
+ * offsets are not advanced past it — until the underlying cause is resolved. This stalling is the expected
+ * behaviour, not a bug: skipping the record would silently diverge the replica from the source journal.
+ */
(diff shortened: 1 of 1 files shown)


replicator/src/main/scala/com/evolution/kafka/journal/replicator/ConsumeTopic.scala
```diff
@@ -13,6 +13,14 @@ import com.evolutiongaming.skafka.consumer.{RebalanceCallback, RebalanceListener
...
 import scala.concurrent.duration.*
 
+/** Consumes a topic and replicates its records into the eventual store.
+  *
+  * By design there is no skip or dead-letter path. The replicator must preserve the exact order and completeness of
+  * the journal, so no record may ever be dropped. A record that cannot be processed (for example a corrupt or
+  * unparseable one) therefore intentionally stalls replication of its topic — the poll is retried with backoff and
+  * offsets are not advanced past it — until the underlying cause is resolved. This stalling is the expected
+  * behaviour, not a bug: skipping the record would silently diverge the replica from the source journal.
+  */
```
mr-git:
```suggestion
/** Consumes a topic and replicates its records into the eventual store.
 *
 * By design there is no skip or dead-letter path. The replicator must preserve the exact order and completeness of
 * the journal, so no record may ever be dropped. A record that cannot be processed (for example a corrupt or
 * unparseable one) therefore intentionally stalls replication of its topic — the poll is retried with backoff and
 * offsets are not advanced past it — until the underlying cause is resolved. This stalling is the expected
 * behaviour, not a bug: skipping the record would silently diverge the replica from the source journal.
 */
```

replicator/src/main/scala/com/evolution/kafka/journal/replicator/ConsumeTopic.scala
```diff
@@ -13,6 +13,14 @@ import com.evolutiongaming.skafka.consumer.{RebalanceCallback, RebalanceListener
...
 import scala.concurrent.duration.*
 
+/** Consumes a topic and replicates its records into the eventual store.
+ *
+ * By design there is no skip or dead-letter path. The replicator must preserve the exact order and completeness of
+ * the journal, so no record may ever be dropped. A record that cannot be processed (for example a corrupt or
+ * unparseable one) therefore intentionally stalls replication of its topic — the poll is retried with backoff and
+ * offsets are not advanced past it — until the underlying cause is resolved. This stalling is the expected
+ * behaviour, not a bug: skipping the record would silently diverge the replica from the source journal.
+ */
```
mr-git:
```suggestion
/**
 * Consumes a topic and replicates its records into the eventual store.
 *
 * By design there is no skip or dead-letter path. The replicator must preserve the exact order and
 * completeness of the journal, so no record may ever be dropped. A record that cannot be processed
 * (for example a corrupt or unparseable one) therefore intentionally stalls replication of its
 * topic — the poll is retried with backoff and offsets are not advanced past it — until the
 * underlying cause is resolved. This stalling is the expected behaviour, not a bug: skipping the
 * record would silently diverge the replica from the source journal.
 */
```

Issue #922, opened 2026-07-19
Poison record permanently stalls HeadCache consumption (TopicCache)


stasimus:
## Problem

The HeadCache background consumption has the same "poison record stalls the topic" shape as the replicator (separate issue), but in a different subsystem.

In `TopicCache` the pipeline is `HeadCacheConsumption.apply(...).foreach { body }.onError{log}.background`. `Stream.around(retry)` (`HeadCacheConsumption.scala:100`) wraps the entire downstream fold — consumer acquisition, seek, poll, **and the `.foreach` body**. So an error raised while *processing* a polled record restarts the whole stream (re-seek to the cache's current offsets, re-poll). Because the poison record was never added, the cache offset for its partition never advances, so the re-poll returns the same record and fails again — an infinite no-progress loop (`TopicCache.scala:145-216`, raise sites `TopicCache.scala:173` and `conversions/ConsRecordToActionHeader.scala:57`).

Two concrete triggers:
- A record whose `journal.action` header is present but unparseable (version/format skew, corruption) → `ConsRecordToActionHeader` raises `JournalError`.
- **Topic repartitioning after cache init**: `HeadCacheConsumption.seek` re-assigns all current partitions each retry, but `cachesMap` is fixed at init. A record on a newly-added partition → `cachesMap.get(partition) == None` → `JournalError("invalid partition")` → same loop.

## Failure scenario

For every journal in the affected topic, `HeadCache.get` returns Timeout/None, so `Journals.headAndStream` falls back to **full Kafka replay on every recovery, indefinitely**. It does not self-heal (`ExpiringCache` uses `expireAfterRead`, reset by continuing reads). Partly acknowledged by TODOs at `HeadCache.scala:32-34` and `TopicCache.scala:216`.

## Suggested direction

Same as the replicator issue: bounded retry + skip/quarantine of an unprocessable record so the cache offset can advance; also handle partitions added after init.

stasimus:
Investigated the two `HeadCache` TODOs (`HeadCache.scala:32-34`) referenced here, since they bear directly on this issue.

### TODO "Fail headcache when background tasks failed" — this is the actionable core of #922

`TopicCache.make` spawns three fibers and **discards the outcome of every one**:

- **Consumption** (`TopicCache.scala:145-217`): `HeadCacheConsumption.apply(...).foreach{…}.onError{ log.error }.background` — the `/*TODO headcache: fail head cache*/` marker sits at `:216`.
[...]

PR #933, opened 2026-07-19
Rebuild head cache when a background task fails

stasimus:
Background-task failures in TopicCache were swallowed, leaving the head cache silently non-functional. Now a failed consumption/remove fiber, and the repartition (unknown-partition) case, evict the topic so the next `get` rebuilds it. Implements the "fail headcache" TODO; relates to #922. Test added.


mr-git:
I tried the unit-test on code from main branch - it doesn't fail!

@stasimus, are you sure that we need this change?

PR #934, opened 2026-07-19
Fail schema migration instead of swallowing errors (#923)

### LINKED ISSUE #923: Schema migrations can be left permanently half-applied (MigrateSchema swallows errors)
## Problem

Schema migrations can be left permanently half-applied.

In `MigrateSchema.run` each migration step is executed and the version is then advanced in the same for-comprehension (`cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala:85-91`):

```scala
.foldLeftM(version) { (version, migration) =>
  val version1 = version + 1
  for {
    _ <- migration.execute.first.void.voidError   // swallows ANY error
    _ <- setVersion(version1)                      // runs unconditionally
  } yield version1
}
```

`.voidError` discards *any* error from the migration step, and `setVersion(version1)` then runs regardless. The migration list (`SetupSchema.migrations`) includes critical `CREATE TABLE pointer2`, `DROP TABLE`, and `ALTER TABLE ... ADD` statements.

## Failure scenario

A migration step (e.g. `createTablePointer2`, or an `ALTER ... ADD`) hits a transient timeout. The error is swallowed, the schema version is advanced past that step, so on the next startup `migrate(version)` drops the step and never retries it. The table/column is never created, prepared-statement setup against it later fails, and the service cannot start — with no self-heal, requiring manual DB intervention.

`.voidError` appears intended to tolerate "already exists" on additive `ALTER`, but it is applied uniformly to creational/destructive steps too.

## Suggested direction

Only advance the version when the step actually succeeds; scope error tolerance to the specific idempotent-DDL cases that need it, rather than swallowing all errors uniformly.

stasimus:
Fixes #923. `MigrateSchema` used `.voidError`, swallowing migration errors and advancing the schema version anyway, so a transient failure permanently skipped a step. Now errors propagate and `setVersion` runs only on success, keeping the version truthful. Adds tests. Note: legacy schemas without a version now fail loudly on re-added columns.

### DIFF
--- a/cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
+++ b/cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
@@ -89,7 +89,8 @@ private[journal] object MigrateSchema {
                 .foldLeftM(version) { (version, migration) =>
                   val version1 = version + 1
                   for {
-                    _ <- migration.execute.first.void.voidError
+                    // Errors are purposefully not swallowed to prevent the stored version to advance
+                    _ <- migration.execute.first.void
                     _ <- setVersion(version1)
                   } yield version1
                 }
--- a/eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/SetupSchemaSpec.scala
+++ b/eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/SetupSchemaSpec.scala
@@ -2,6 +2,7 @@ package com.evolution.kafka.journal.eventual.cassandra
 
 import cats.implicits.catsStdInstancesForTry
 import cats.syntax.all.*
+import com.datastax.driver.core.exceptions.InvalidQueryException
 import com.datastax.driver.core.{PreparedStatement, RegularStatement, Row, Statement}
 import com.evolution.kafka.journal.cassandra.CassandraSync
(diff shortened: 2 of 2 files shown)


cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
```diff
@@ -85,7 +85,12 @@ private[journal] object MigrateSchema {
                 .foldLeftM(version) { (version, migration) =>
                   val version1 = version + 1
                   for {
-                    _ <- migration.execute.first.void.voidError
+                    // Any migration failure is propagated: the `foldLeftM` short-circuits before
+                    // `setVersion` runs, so the persisted version is only ever advanced past a step
+                    // that fully succeeded. This keeps the schema version truthful (version == k
+                    // iff steps 0..k were all applied) and lets the next start resume from the last
+                    // good step, instead of silently skipping a half-applied migration. See #923.
```
nviliunov-evolution-throwaway: Suggest to drop the entire comment or to shorten it to one line
```suggestion
                    // Errors are purposefully not swallowed to prevent the stored version to advance
```

mr-git (commented):
Overall I like this, though I'd like to make sure that we have **all** storages in sync and in good shape before we update to version, which contains this change (yes, I am a bit afraid).

mr-git:
blocked by #944!

stasimus:
Re-triggering required check after org policy rollout fix

PR #935, opened 2026-07-19
Reject per-write serialization failures instead of failing store

stasimus:
Per the Akka/Pekko journal contract, a serialization failure for one `AtomicWrite` should surface as a per-write rejection (`Failure` → `onPersistRejected`), not by failing `F`. Failing `F` is treated as a store failure that stops/restarts the persistent actor and drops the other valid writes in the same batch.

### DIFF
--- a/akka/persistence/src/main/scala/com/evolution/kafka/journal/akka/persistence/JournalAdapter.scala
+++ b/akka/persistence/src/main/scala/com/evolution/kafka/journal/akka/persistence/JournalAdapter.scala
@@ -3,7 +3,7 @@ package com.evolution.kafka.journal.akka.persistence
 import akka.persistence.{AtomicWrite, PersistentRepr}
 import cats.effect.*
 import cats.syntax.all.*
-import cats.{Monad, Parallel, ~>}
+import cats.{Monad, MonadThrow, Parallel, ~>}
 import com.evolution.kafka.journal.*
 import com.evolution.kafka.journal.conversions.{ConversionMetrics, KafkaRead, KafkaWrite}
 import com.evolution.kafka.journal.eventual.cassandra.EventualCassandra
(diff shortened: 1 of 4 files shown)


akka/persistence/src/main/scala/com/evolution/kafka/journal/akka/persistence/JournalAdapter.scala
```diff
@@ -163,17 +163,35 @@ object JournalAdapter {
...
+            // Serialize each atomic write independently so that a serialization failure is surfaced as a
+            // per-write rejection (`Failure`, routed to `onPersistRejected`) rather than by failing `F` - which
+            // Akka/Pekko treats as a store failure that stops/restarts the persistent actor, taking down the
+            // unrelated valid writes in the same batch.
+            results <- aws.toList.traverse { aw =>
+              aw.payload.toList.toNel.traverse { prs =>
+                prs.traverse(serializer.toEvent).map { events => (prs, events) }
+              }.attempt
+            }
+            accepted = results.collect { case Right(Some(prsAndEvents)) => prsAndEvents }
```
mr-git: Shouldn't we stop at first failure to serialize?

Like, if we have events: `E0s`, `E1s`, `E2f`, `E3s`, then we could accept events `E0s` and `E1s` for writing, as we can serialize them, but, when we fail to serialize `E2f`, then we should not accept `E3s` any more even while it is serializable, right?

IMHO we should track the result of last serialization and if it is `failure`, then it should gets populated as such state for all following events.

stasimus: Agree, fixed

mr-git:
I am not sure, if we should merge this!

**IFF** I understood the proposed change correctly!!!

Let's imagine the situation where Replicator:
* runs and it gets a batch with 100 records with events for 100 aggregates, where first one is "defective"
* would NOT persist the first one while it would write in storage all other events

The consequences:
* Replicator cannot commit offsets
* technically, Replicator could keep up replication of events for 99 aggregates, but on next redeployment (within Kafka's retention period), it would have to re-consume and re-process them all - this would cause significant delay of real-time activities
* if next redeployment happens after broken record has been dropped by Kafka's retention policy, Replicator could persist events, which came after broken record - thus it would create the "gap", which could introduce inconsistency

Currently, Replicator would fail fast and we have clear visibility where and why. But with this PR, Replication would continue silently (or some metric, which is NOT exposed in this PR, could trigger the alert). It might happen, that we would notice the problem many days or weeks after problem started.

Issue #925, opened 2026-07-19
Unbounded parallelism in replication hot loop (TopicReplicator parFoldMap1)


stasimus:
## Problem

Unbounded parallelism in the replication hot loop.

`TopicReplicator` fans out with `parFoldMap1` over partitions and over keys (`replicator/src/main/scala/com/evolution/kafka/journal/replicator/TopicReplicator.scala:123` and `:150`):

```scala
records.parFoldMap1 { ... }                 // over partitions
records.groupBy(key).parFoldMap1 { ... }    // over keys
```

`parFoldMap1` (cats-helper `ParallelHelper`) uses `Parallel`, i.e. for `IO` it starts **all** elements concurrently — there is no concurrency bound.

## Failure scenario

A single poll that spans thousands of distinct keys spawns thousands of concurrent Cassandra write chains at once. This can exhaust the driver's max in-flight-requests-per-connection (`BusyPoolException` / `NoHostAvailable`), spike latency, or OOM under large fetch sizes. The failing poll then feeds directly into the poison-record retry loop (separate issue), amplifying the outage.

## Suggested direction

Bound the fan-out (e.g. `parReplicateAN` / a semaphore / a configurable parallelism limit) sized to the Cassandra connection pool.

mr-git:
With default `max-poll-records = 1000` it is unlikely that Replicator will get 1000 different keys (mostly because lot more aggregates are active at any time, usually each aggregate tends to produce several events in its own "iteration" and Kafka tends to batch records per topic-partition).

In addition, Cats Effect has its own parallelism control in the form of [computeWorkerThreadCount](https://github.com/typelevel/cats-effect/blob/18fdf52498b00f9908caea77b8c132664656c781/core/jvm/src/main/scala/cats/effect/IOApp.scala#L199) which limits how many threads are actually active at any time (usually we have this number well below 10).

IMHO this sounds like a theoretical issue - we could introduce the section about performance tuning in README, but that will require a lot of work to define and describe the tuning process beyond current defaults.

PR #936, opened 2026-07-19
Bound TopicReplicator key replication concurrency

stasimus:
The key-level `parFoldMap1` in TopicReplicator started a Cassandra write chain per key with no limit, so a poll spanning thousands of keys could exhaust the connection pool. A per-topic semaphore now bounds concurrent key replication, sized by the new `replication-parallelism` config (default 100).

### DIFF
--- a/replicator/src/main/scala/com/evolution/kafka/journal/replicator/TopicReplicator.scala
+++ b/replicator/src/main/scala/com/evolution/kafka/journal/replicator/TopicReplicator.scala
@@ -4,6 +4,7 @@ import cats.Monad
 import cats.data.{NonEmptyList as Nel, NonEmptyMap as Nem, NonEmptySet as Nes}
 import cats.effect.*
 import cats.effect.implicits.*
+import cats.effect.std.Semaphore
 import cats.syntax.all.*
 import com.evolution.kafka.journal.*
 import com.evolution.kafka.journal.conversions.{ConsRecordToActionRecord, KafkaRead}
@@ -39,6 +40,7 @@ private[journal] object TopicReplicator {
     metrics: TopicReplicatorMetrics[F],
     cacheOf: CacheOf[F],
     replicatedOffsetNotifier: ReplicatedOffsetNotifier[F],
+    replicationParallelism: Int,
   ): Resource[F, F[Outcome[F, Throwable, Unit]]] = {
 
     implicit val fromAttempt: FromAttempt[F] = FromAttempt.lift[F]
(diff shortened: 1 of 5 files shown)


mr-git:
Is this somehow related to https://github.com/evolution-gaming/kafka-journal/issues/925? If yes, then: https://github.com/evolution-gaming/kafka-journal/issues/925#issuecomment-5024208741.

According to AI, Java driver for Cassandra (both v3 and v4) by default sets `requests per connection` to 1024. In `scassandra`, we have set it to `32768` for `LOCAL` and `2000` for `REMOTE` connections (see: https://github.com/evolution-gaming/scassandra/blob/e86effae8156496a1485d2014b98685220c7b13e/scassandra/src/main/scala/com/evolutiongaming/scassandra/PoolingConfig.scala#L81-L93)- well above proposed `100`.

stasimus:
Good point — raised the default to `1024` to match the Cassandra driver's default requests-per-connection, keeping the bound as a safety cap well below scassandra's configured pool (32768 LOCAL / 2000 REMOTE) while still preventing unbounded fan-out. It's configurable via `replication-parallelism` for tuning.

replicator/src/main/scala/com/evolution/kafka/journal/replicator/TopicReplicator.scala
```diff
@@ -104,6 +108,9 @@ private[journal] object TopicReplicator {
         for {
           journal <- journal.journal(topic)
           cache <- cacheOf[Partition, PartitionFlow](topic)
+          // bounds concurrent per-key replication across all partitions of the topic, so a single
+          // poll spanning many keys cannot exhaust the Cassandra connection pool
+          semaphore <- Semaphore[F](replicationParallelism.max(1).toLong).toResource
```
mr-git: The `.max(1)` feels like we do not trust our own internal `config` :/ 

Wouldn't it be better to move the `.max(1)` to [ReplicatorConfig.scala‎:82](https://github.com/evolution-gaming/kafka-journal/pull/936/changes#diff-52ada19c2d8300e4b2bc8d8cfb0b4bf97368696ac1707818e4253e37559bee61)? And, while we are there, maybe it is worth to change the type to `Long` too?

mr-git (commented):
Overall, I do not mind this change, though we will have to check if it somehow manifests during load-testing.

Probably we will have to find the minimal value, at which it starts to affect the performance of Replicator under load - we have to know how our service behaves.

Issue #16, opened 2018-06-13
Minimise project external dependencies 


t3hnar:

stasimus:
Addressed the only concrete opportunities remaining (unused `scala-java8-compat` and redundant transitive `hostname` in the journal module) in #939. All other dependencies map to active functionality. Closing as stale/unbounded — dependency hygiene is better handled per-PR going forward.

PR #939, opened 2026-07-20
Drop unused journal dependencies

stasimus:
Removes two redundant `journal` dependency declarations flagged by #16.

- `scala-java8-compat`: no source usage; `.asJava` calls use stdlib `scala.jdk.CollectionConverters`.
- `hostname`: only used in `core/HostName.scala` and pulled transitively via `journal dependsOn core`.

Compiles on Scala 2.13 and 3.3.


build.sbt
```diff
@@ -148,9 +148,7 @@ lazy val journal = project
       CatsHelper,
       PlayJson,
       PlayJsonJsoniter,
-      Hostname,
```
mr-git, line 151: module still gets it transitively from `core`

mr-git (approved):

PR #940, opened 2026-07-20
Package replicator as standalone runnable jar

### LINKED ISSUE #15: package replicator as standalone runnable jar

stasimus:
Closes #15. Adds an `sbt-assembly` fat-jar build for the replicator via a new non-published `replicator-app` module with an `IOApp` entrypoint and logback config. Build: `sbt replicatorApp/assembly`, run: `java -jar <jar>`.


mr-git:
I wonder, why would we need it? Why not just close #15?

mr-git:
thanks, but no, I will not maintain it :)

PR #944, opened 2026-07-21
expose schema version (index of last applied migration) in Replicator's metrics

mr-git:
blocks #934!

### DIFF
--- a/cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
+++ b/cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
@@ -31,7 +31,7 @@ private[journal] trait MigrateSchema[F[_]] {
     fresh: MigrateSchema.Fresh,
   )(implicit
     session: CassandraSession[F],
-  ): F[Unit]
+  ): F[Int]
 
 }
 
@@ -51,7 +51,8 @@ private[journal] object MigrateSchema {
    *   schemas, to ensure there is no accidental overwrite if both schemas are located in one
    *   keyspace.
    * @param migrations
-   *   List of CQL statements to execute. The schema version is equal to the size of this list.
+   *   List of CQL statements to execute. The schema version is equal to the index of the last
+   *   migration in this list.
    * @return
    *   The instance of schema migrator.
    */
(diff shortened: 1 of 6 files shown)


cassandra/src/main/scala/com/evolution/kafka/journal/cassandra/MigrateSchema.scala
```diff
@@ -89,7 +93,6 @@ private[journal] object MigrateSchema {
                     _ <- setVersion(version1)
```
stasimus, line 93: `.voidError` swallows migration failures but `setVersion` still advances, so the new version metric reports failed migrations as applied.

mr-git: yes, I want to expose the current "version" before I merge https://github.com/evolution-gaming/kafka-journal/pull/934

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/SetupSchemaSpec.scala
```diff
@@ -194,7 +194,7 @@ class SetupSchemaSpec extends AnyFunSuite with Matchers {
     }
   }
 
-  def migrate(fresh: Boolean): StateT[Unit] = {
+  def migrate(fresh: Boolean): StateT[Int] = {
```
stasimus, line 197: IMHO the returned version is never asserted; every test binds it to `_`, so the value this PR exists to compute has no coverage.

mr-git: will improve!

journal/src/main/scala/com/evolution/kafka/journal/eventual/ReplicatedJournal.scala
```diff
@@ -213,6 +217,12 @@ object ReplicatedJournal {
         labels = LabelNames("version"),
       )
 
+      val schemaVersionInfo = registry.info(
```
stasimus, line 220: A gauge with `.set(version)` fits a numeric version better than an info metric, which pins the value to 1 and leaves a stale series per version across restarts.

stasimus: IMHO

stasimus:
Overall is better than was

journal/src/main/scala/com/evolution/kafka/journal/eventual/ReplicatedJournal.scala
mr-git: The stale metric lasts only for 5 minutes (depends on Prometheus config). I decided to use `info` approach as it feels better fit for all kinds of versions:
```
replicated_journal_info{version="9.1.1+17-010968a2+20260721-1650"} 1
replicated_journal_schema_version_info{version="8"} 1
replicator_info{version="Unknown"} 1
replicator_kafka_journal_client_version_info{topic="journal.replicator-test",version="9.1.1+17-010968a2+20260721-1650"}
```

PR #950, opened 2026-07-22
Bump sstream 1.2.0, fuse consume loop, add benchmarks

stasimus:
Bump sstream 1.1.0→1.2.0: source-compatible, all tests green, read path neutral-to-faster. Fuse `ConsumeActionRecords` poll loop into one traverse: ~3.9% allocations/op. Add `StreamActionRecordsBenchmark` and `ConsumeActionRecordsBenchmark` covering the sstream read/consume paths; fix `BatchBenchmark` eta-expansion so the module compiles.

### DIFF
--- a/benchmark/src/test/scala/com/evolution/kafka/journal/ConsumeActionRecordsBenchmark.scala
+++ b/benchmark/src/test/scala/com/evolution/kafka/journal/ConsumeActionRecordsBenchmark.scala
@@ -1,0 +1,80 @@
+package com.evolution.kafka.journal
+
+import cats.data.NonEmptyList as Nel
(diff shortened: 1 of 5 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,137 @@
+{
+  "permissions": {
+    "allow": [
```
mr-git: is this something required?

IMHO the `.claude` should go into `.gitignore`, unless it provides per-project "hints"

mr-git:
sstream `1.2.0` contains concurrency issue and will **NOT** be merged!

PR #956, opened 2026-07-29
Revert "Bound TopicReplicator key replication concurrency" (#936)

stasimus:
The per-topic Semaphore causes a CPU storm: any key replication error cancels all sibling fibers, each running O(n) waiter-queue cleanup in a CAS loop. Reverting as immediate mitigation


stasimus:
Revert due to https://github.com/typelevel/cats-effect/issues/4434

PR #963, opened 2026-08-07
journal brain split / fork detection mechanism

rtar:
Not ready for review yet

### DIFF
--- a/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
+++ b/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
@@ -1,7 +1,7 @@
 package com.evolution.kafka.journal.eventual.cassandra
 
 import cats.syntax.all.*
-import com.evolution.kafka.journal.{DeleteTo, EventRecord, Key, Origin, PartitionOffset, SeqNr}
+import com.evolution.kafka.journal.{EventRecord, Key, Origin, PartitionOffset, SeqNr}
 
 /**
  * A candidate journal fork: an event whose `seqNr` is not above every `seqNr` replicated to that
@@ -10,16 +10,14 @@ import com.evolution.kafka.journal.{DeleteTo, EventRecord, Key, Origin, Partitio
  * Happens when a persistent actor's append to Kafka is still in flight while the entity is
  * restarted elsewhere: the new incarnation does not see the in-flight event, so it appends a
  * different event with the same `seqNr`. Both are replicated (the `journal` table clusters on
- * `(seq_nr, timestamp)`), and the next recovery of the entity fails with `Data integrity violated:
+ * `(seq_nr, timestamp)`), and a recovery of the entity can then fail with `Data integrity violated:
  * seqNr ... duplicated in multiple records`, see
  * [[com.evolution.kafka.journal.eventual.cassandra.EventualCassandra]].
  *
- * Whether the `seqNr` really is written twice is only sometimes decidable here, see
- * [[JournalFork.Consequence]]: a recovery fails on two rows sharing a `seq_nr`, and knowing which
- * `seqNr`s a journal already occupies would take a `journal` table read this deliberately avoids.
- *
- * Detected, but not acted upon: which of the two branches survives is a decision for a human or a
- * repair tool, as it depends on the `writerUuid` of the events which follow the fork.
+ * Detected, but not acted upon, and deliberately says nothing about what it costs: that depends on
+ * how the journal is used, which is not known here. Which of the two branches survives is a
+ * decision for a human or a repair tool, as it depends on the `writerUuid` of the events which
+ * follow.
  *
  * The two records are named by their Kafka offset order, because that is all they always have in
  * common: `earlierRecord` may be the journal's last replicated event, but it may equally be another
@@ -31,10 +29,7 @@ import com.evolution.kafka.journal.{DeleteTo, EventRecord, Key, Origin, Partitio
  *   the event being appended right now, the one whose `seqNr` failed to increase
  * @param earlierRecord
  *   the record with the highest `seqNr` at a lower offset - either the journal's last replicated
- *   event, see [[JournalFork.Record.fromJournalHead]], or an earlier event of the same batch. It is
- *   the record actually duplicated only when `duplicateProven` came from comparing against it.
- * @param deleteTo
- *   the journal's delete watermark at the time
+ *   event, see [[JournalFork.Record.fromJournalHead]], or an earlier event of the same batch.
  * @param duplicateProven
  *   whether a record is known to occupy `seqNr` already - true when `laterRecord` repeats the
  *   `seqNr` of the journal head or of an earlier event of the same batch. False means the `seqNr`
@@ -45,24 +40,10 @@ private[journal] final case class JournalFork(
   key: Key,
   laterRecord: JournalFork.Record,
   earlierRecord: JournalFork.Record,
-  deleteTo: Option[DeleteTo],
   duplicateProven: Boolean,
 ) {
 
-  /**
-   * The `seqNr` which failed to increase.
-   */
   def seqNr: SeqNr = laterRecord.seqNr
-
-  /**
-   * A `seqNr` at or below `deleteTo` is harmless whether it is duplicated or not: the record it
-   * would duplicate is gone from the `journal` table and [[EventualCassandra]] reads above it.
-   */
-  def consequence: JournalFork.Consequence = {
-    if (deleteTo.exists(_.value >= seqNr)) JournalFork.Consequence.BelowDeleteTo
-    else if (duplicateProven) JournalFork.Consequence.BreaksRecovery
-    else JournalFork.Consequence.SuspectedRegression
-  }
 
   def show: String = {
     s"key: $key, later: ${ laterRecord.show }, earlier: ${ earlierRecord.show }"
@@ -84,7 +65,6 @@ private[journal] object JournalFork {
     events: List[EventRecord[A]],
   ): List[JournalFork] = {
 
-    val deleteTo = journalHead.flatMap(_.deleteTo)
     val headSeqNr = journalHead.map(_.seqNr)
 
     // `seqNr`s of the batch walked so far: needed on top of `earlier` because two *equal* `seqNr`s
@@ -101,7 +81,6 @@ private[journal] object JournalFork {
               key,
               laterRecord = record,
               earlierRecord = earlier,
-              deleteTo = deleteTo,
               duplicateProven = seen.contains(record.seqNr) || headSeqNr.contains(record.seqNr),
             )
             (Some(earlier), seen1, fork :: forks)
@@ -114,47 +93,7 @@ private[journal] object JournalFork {
   }
 
   /**
-   * What a [[JournalFork]] means for the journal, and hence how loudly to report it. The `name` is
-   * a metric label value, so the set of them stays small and stable.
-   */
-  sealed abstract class Consequence(val name: String, val explanation: String)
-
-  object Consequence {
-
-    /**
-     * A record is known to occupy the `seqNr` and is still readable, so the entity's next recovery
-     * fails and stays broken until the losing row is removed by hand. Assumes the default
-     * `seqNrUniqueness` data integrity setting on the reader side; with it off,
-     * [[EventualCassandra]] replays both events instead of failing.
-     */
-    case object BreaksRecovery
-    extends Consequence("breaks_recovery", "the next recovery of the entity will fail on it")
-
-    /**
-     * The `seqNr` is at or below the journal's `deleteTo`, so whatever it duplicates is already
-     * gone and no recovery reads that far back. The appended row is orphaned until the next purge
-     * or expiry - the same underlying bug, but nothing to repair.
-     */
-    case object BelowDeleteTo
-    extends Consequence(
-      "below_delete_to",
-      "already deleted, so no recovery reads it - the appended row is merely orphaned",
-    )
-
-    /**
-     * The `seqNr` regressed but nothing proves it is occupied: it is the fingerprint of the fork
-     * bug, yet concurrent appends of distinct `seqNr`s to one key look identical from here.
-     */
-    case object SuspectedRegression
-    extends Consequence(
-      "suspected_regression",
-      "a duplicate is likely but unproven - only a `journal` table read for this seqNr can tell",
-    )
-  }
-
-  /**
-   * Where one of the two records claiming the same `seqNr` sits in Kafka, and which node appended
-   * it.
+   * Where one of a fork's two records sits in Kafka, and which node appended it.
    *
    * Notably not the `writerUuid` of the entity incarnation behind it, which is what decides the
    * surviving branch - that lives in the payload and is left to the repair tool.
--- a/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
+++ b/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
@@ -61,7 +61,7 @@ private[journal] final case class JournalFork(
    * from the `journal` table and [[EventualCassandra]] starts reading above `deleteTo`.
    */
   def consequence: JournalFork.Consequence = {
-    if (deleteTo.exists { _.value >= seqNr }) JournalFork.Consequence.BelowDeleteTo
+    if (deleteTo.exists(_.value >= seqNr)) JournalFork.Consequence.BelowDeleteTo
     else if (duplicateProven) JournalFork.Consequence.BreaksRecovery
     else JournalFork.Consequence.SuspectedRegression
   }
@@ -99,8 +99,8 @@ private[journal] object JournalFork {
     events: List[EventRecord[A]],
   ): List[JournalFork] = {
 
-    val deleteTo = journalHead.flatMap { _.deleteTo }
-    val headSeqNr = journalHead.map { _.seqNr }
+    val deleteTo = journalHead.flatMap(_.deleteTo)
+    val headSeqNr = journalHead.map(_.seqNr)
 
     // `seqNr`s of the batch walked so far: needed on top of `earlier` because two *equal* `seqNr`s
     // both below the running maximum would otherwise never be compared to each other
--- a/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
+++ b/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
@@ -119,9 +119,9 @@ private[journal] object JournalFork {
               deleteTo = deleteTo,
               duplicateProven = seen.contains(record.seqNr) || headSeqNr.contains(record.seqNr),
             )
-            (earlier.some, seen1, fork :: forks)
+            (Some(earlier), seen1, fork :: forks)
           case _ =>
-            (record.some, seen1, forks)
+            (Some(record), seen1, forks)
         }
     }
 
--- a/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
+++ b/eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
@@ -31,9 +31,8 @@ import com.evolution.kafka.journal.{DeleteTo, EventRecord, Key, Origin, Partitio
  *   the event being appended right now, the one whose `seqNr` failed to increase
  * @param earlierRecord
  *   the record with the highest `seqNr` at a lower offset - either the journal's last replicated
- *   event, see [[JournalFork.Record.fromJournalHead]], or an earlier event of this very batch,
- *   which is not in Cassandra yet at this point. This is the invariant that was violated; it is the
- *   record actually duplicated only when `duplicateProven` was established by comparing against it.
+ *   event, see [[JournalFork.Record.fromJournalHead]], or an earlier event of the same batch. It is
+ *   the record actually duplicated only when `duplicateProven` came from comparing against it.
  * @param deleteTo
  *   the journal's delete watermark at the time
  * @param duplicateProven
@@ -51,14 +50,13 @@ private[journal] final case class JournalFork(
 ) {
 
   /**
-   * The `seqNr` at stake, i.e. the one which failed to increase.
+   * The `seqNr` which failed to increase.
    */
   def seqNr: SeqNr = laterRecord.seqNr
 
   /**
-   * What this means for the journal, see [[JournalFork.Consequence]]. A `seqNr` at or below
-   * `deleteTo` is harmless whether it is duplicated or not: the record it would duplicate is gone
-   * from the `journal` table and [[EventualCassandra]] starts reading above `deleteTo`.
+   * A `seqNr` at or below `deleteTo` is harmless whether it is duplicated or not: the record it
+   * would duplicate is gone from the `journal` table and [[EventualCassandra]] reads above it.
    */
   def consequence: JournalFork.Consequence = {
     if (deleteTo.exists(_.value >= seqNr)) JournalFork.Consequence.BelowDeleteTo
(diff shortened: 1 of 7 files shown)


mr-git:
@rtar , could you please describe what is an idea or planned strategy?

I'd love to see some documentation in README on where to look, what to do etc

journal/src/main/scala/com/evolution/kafka/journal/eventual/ReplicatedJournal.scala
```diff
@@ -178,6 +178,17 @@ object ReplicatedJournal {
...
+
+    /**
+     * Accounts a journal fork, i.e. an event replicated with a `seqNr` which did not increase over
+     * what the journal already had.
+     *
+     * @param duplicateProven
+     *   whether a record is known to occupy the `seqNr` already, or it merely failed to increase.
+     *   Deliberately a fact rather than a severity - what a fork costs depends on how the journal
+     *   is used, which is not known here.
+     */
```
mr-git, line 190: this needs rewording, maybe something like (requires `fmt`!):
```suggestion
     * @param duplicateProven
     *   indicates whether a record (`kind`):
     *    - (`duplicate`) is known to use the `seqNr` which was used previously
     *    - (`regression`) unexpectedly failed to increase `seqNr`, could mean re-replicated older event
     *
     *   Deliberately a fact rather than a severity - it is not known here what are the consequences a forked journal - it depends on how the journal is used
     */
```

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
```diff
@@ -0,0 +1,125 @@
...
+ * journal before it, which the `seqNr`s of a journal are supposed to be.
+ *
+ * Happens when a persistent actor's append to Kafka is still in flight while the entity is
+ * restarted elsewhere: the new incarnation does not see the in-flight event, so it appends a
+ * different event with the same `seqNr`. Both are replicated (the `journal` table clusters on
+ * `(seq_nr, timestamp)`), and a recovery of the entity can then fail with `Data integrity violated:
+ * seqNr ... duplicated in multiple records`, see
+ * [[com.evolution.kafka.journal.eventual.cassandra.EventualCassandra]].
+ *
+ * Detected, but not acted upon, and deliberately says nothing about what it costs: that depends on
```
mr-git, line 17: not sure about "what it costs"...  IMHO it is more about "what are the consequences"

mr-git: a bit annoying that practically the same text is used just for `duplicateProven` field in metrics, probably the code in metrics at least should reference this class

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
```diff
@@ -0,0 +1,125 @@
...
+ * @param laterRecord
+ *   the event being appended right now, the one whose `seqNr` failed to increase
+ * @param earlierRecord
+ *   the record with the highest `seqNr` at a lower offset - either the journal's last replicated
+ *   event, see [[JournalFork.Record.fromJournalHead]], or an earlier event of the same batch.
+ * @param duplicateProven
+ *   whether a record is known to occupy `seqNr` already - true when `laterRecord` repeats the
+ *   `seqNr` of the journal head or of an earlier event of the same batch. False means the `seqNr`
+ *   merely regressed: a duplicate is likely, but concurrent appends of *distinct* `seqNr`s to one
+ *   key regress too, so it is not proof.
```
mr-git, line 37: requires `fmt`!!!
```suggestion
 *   whether a record is known to occupy `seqNr` already:
 *    - `true` when `laterRecord` repeats the `seqNr` of the journal head or of an earlier event of the same batch
 *    - `false` means the `seqNr` merely regressed: a duplicate is likely, but concurrent appends of *distinct* `seqNr`s to one key regress too, so it is not proof
```

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
```diff
@@ -0,0 +1,125 @@
...
+ * [[com.evolution.kafka.journal.eventual.cassandra.EventualCassandra]].
+ *
+ * Detected, but not acted upon, and deliberately says nothing about what it costs: that depends on
+ * how the journal is used, which is not known here. Which of the two branches survives is a
+ * decision for a human or a repair tool, as it depends on the `writerUuid` of the events which
+ * follow.
+ *
+ * The two records are named by their Kafka offset order, because that is all they always have in
+ * common: `earlierRecord` may be the journal's last replicated event, but it may equally be another
+ * event of the same batch, still being appended alongside `laterRecord`.
```
mr-git, line 24: about "all they always have in common" part - the `offset` is the different part, they have in common the `key` and `seqNr` (also, most probably, the `writerUuid` is different too)

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @param key
+ *   the journal
+ * @param laterRecord
+ *   the event being appended right now, the one whose `seqNr` failed to increase
+ * @param earlierRecord
```
mr-git, line 30: the names need some reconsideration - even with "docs" they do not explain a lot

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
mr-git, line 24:
```suggestion
 * event of the same batch, still being appended alongside `laterRecord` (like events produced by different writers, consumed during single `poll`).
```

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalFork.scala
```diff
@@ -0,0 +1,125 @@
...
+    val headSeqNr = journalHead.map(_.seqNr)
+
+    // `seqNr`s of the batch walked so far: needed on top of `earlier` because two *equal* `seqNr`s
+    // both below the running maximum would otherwise never be compared to each other
+    val (_, _, forks) = events.foldLeft(
+      (journalHead.map(Record.fromJournalHead), Set.empty[SeqNr], List.empty[JournalFork]),
+    ) {
+      case ((earlier, seen, forks), event) =>
+        val record = Record.fromEventRecord(event)
+        val seen1 = seen + record.seqNr
```
mr-git: I worry, if this could produce too much garbage - we might have thousands of events per batch and here we will produce new and larger `Set` on every event

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkReporter.scala
```diff
@@ -0,0 +1,44 @@
...
+  def fromMetrics[F[_]: Applicative](
+    metrics: ReplicatedJournal.Metrics[F],
+    log: Log[F],
+  ): JournalForkReporter[F] = { fork =>
+    // the headline repeats the `seqNr ... duplicated` wording of the recovery side error, so that a
+    // single grep finds both the fork and the recoveries it later breaks - but only where the
+    // duplicate is proven, so that the phrase keeps meaning what it says
+    val headline =
+      if (fork.duplicateProven) s"Data integrity violated: seqNr ${ fork.seqNr } duplicated by a journal fork"
+      else s"Suspected journal fork: seqNr ${ fork.seqNr } did not increase"
```
mr-git, line 36: should we document, that `suspected journal fork` can happen when events were fully replicated, but Replicator didn't have chance or failed to commit the offsets?

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
```diff
@@ -139,6 +156,28 @@ private[journal] object ReplicatedCassandra {
...
+                          } { offset =>
+                            events.filter { event => event.partitionOffset.offset > offset }
+                          }
+                        }
+
+                        /**
+                         * Free of extra Cassandra reads: `journalHead` is the one already loaded
+                         * for the append itself.
+                         */
+                        def detectForks(journalHead: Option[JournalHead]) = {
```
mr-git, line 174: current version will always attempt to detect forks even if metrics (and reporting?) are disabled, right?

do we need a way to enable/disable it?

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  test("a seqNr gap is not a fork") {
+    val events = List(event(seqNr = 9, offset = 3))
+    val forks = JournalFork.fromEvents(key, journalHead(seqNr = 2, offset = 2), events)
+
+    assert(forks.isEmpty)
+  }
+
+  test("repeating the journal head seqNr proves a duplicate") {
```
mr-git, line 64: do we need a unit-test for the case when:
* do `poll`
* replicate event
* fail to commit
* repeat the `poll` after recovery?

if the "new event" has the same `offset` as `journalHead`, then it is the same event and must not trigger "fork" case, right?

something like:
```scala
  test("re-replicating the journal head doesn't trigger fork detector") {
    val events = List(event(seqNr = 2, offset = 2))
    val forks = JournalFork.fromEvents(key, journalHead(seqNr = 2, offset = 2), events)

    assert(forks.isEmpty)
  }
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    val events = List(event(seqNr = 2, offset = 3))
+    val forks = JournalFork.fromEvents(key, journalHead(seqNr = 2, offset = 2), events)
+
+    assert(forks.map(_.seqNr.value) == List(2L))
+    assert(forks.map(_.duplicateProven) == List(true))
+  }
+
+  test("a bare regression is only suspected - concurrent appends of distinct seqNrs look the same") {
+    // what `JournalPerfSpec.appendNoise` does: distinct seqNrs appended to one key in parallel, so
+    // they arrive out of order without any of them being written twice
```
mr-git, line 74: "poison" from somewhere else?
```suggestion
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    val events1 = List(event(seqNr = 2, offset = 6))
+    val events2 = List(event(seqNr = 3, offset = 1), event(seqNr = 1, offset = 2), event(seqNr = 2, offset = 3))
+    val forks1 = JournalFork.fromEvents(key, journalHead(seqNr = 5, offset = 5), events1)
+    val forks2 = JournalFork.fromEvents(key, none, events2)
+
+    assert(forks1.map(_.duplicateProven) == List(false))
+    assert(forks2.map(_.duplicateProven) == List(false, false))
```
mr-git, line 81: why 2 slightly different cases are interleaved here? the grouping of similar setup doesn't help with reading!

please consider to split the case in two dedicated:
```
val events1 = List(event(seqNr = 2, offset = 6)) // `seqNr` before the "head" 
val forks1 = JournalFork.fromEvents(key, journalHead(seqNr = 5, offset = 5), events1)
assert(forks1.map(_.duplicateProven) == List(false))

// similar, but different case!
// "head" set at `seqNr=3`, next events are "before" the "head"
val events2 = List(event(seqNr = 3, offset = 1), event(seqNr = 1, offset = 2), event(seqNr = 2, offset = 3))
val forks2 = JournalFork.fromEvents(key, none, events2)
assert(forks2.map(_.duplicateProven) == List(false, false))
```

I am not sure about value of `duplicateProven` in  second case - technically it is a fork from "blank" state (we allow gaps in sequence numbers) 🤔

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  test("every fork of a batch is reported, in event order") {
+    val events = List(event(seqNr = 2, offset = 4), event(seqNr = 3, offset = 5), event(seqNr = 1, offset = 6))
```
mr-git, line 85: why do we need extra `seqNr=1` case? it duplicates the `seqNr=2` by its _shape_ (`seqNr` is before with `offset` "after" the head)
```suggestion
    // `seqNr=2` is before "head"
    // `seqNr=3` is different version of "head"
    // `seqNr=1` is before "head"
    val events = List(event(seqNr = 2, offset = 4), event(seqNr = 3, offset = 5), event(seqNr = 1, offset = 6))
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    val forks = JournalFork.fromEvents(key, journalHead(seqNr = 3, offset = 3), events)
+
+    assert(forks.map(_.seqNr.value) == List(2L, 3L, 1L))
+  }
+
+  test("a fork does not advance the earlier seqNr") {
+    // seqNr 4 follows the head, so it is not a fork itself, but it does become what the events after
+    // it are compared against
```
mr-git, line 93: the comment is very hard to comprehend, maybe (requires `fmt`!)?:
```suggestion
    // first event with `seqNr=4` follows the head, so it is not a fork itself, but last event with `seqNr=4` creates a "fork"
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    // them mere regressions - they are compared against the seqNrs already seen instead
+    val events = List(event(seqNr = 9, offset = 1), event(seqNr = 1, offset = 2), event(seqNr = 1, offset = 3))
+    val forks = JournalFork.fromEvents(key, none, events)
+
+    assert(forks.map(_.duplicateProven) == List(false, true))
+  }
+
+  test("origins of both records are carried over where known") {
+    val later = event(seqNr = 1, offset = 2, origin = Origin("later"))
+    val earlier = event(seqNr = 1, offset = 1, origin = Origin("earlier"))
```
mr-git, line 119: please define them "chronologically":

```suggestion
    val earlier = event(seqNr = 1, offset = 1, origin = Origin("earlier"))
    val later = event(seqNr = 1, offset = 2, origin = Origin("later"))
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/JournalForkTest.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    val withinBatch = JournalFork.fromEvents(key, none, List(earlier, later))
+    val againstHead = JournalFork.fromEvents(key, journalHead(seqNr = 1, offset = 1), List(later))
+
+    assert(withinBatch.map(_.laterRecord.origin) == List(Some(Origin("later"))))
+    assert(withinBatch.map(_.earlierRecord.origin) == List(Some(Origin("earlier"))))
+
+    // against the journal head the origin of the earlier writer is not at hand, see
+    // `JournalFork.Record.fromJournalHead`
+    assert(againstHead.map(_.earlierRecord.origin) == List(none))
```
mr-git, line 129: IMHO, keeping each "flow" together makes more then sense than grouping by "activity"
```suggestion
    assert(withinBatch.map(_.laterRecord.origin) == List(Some(Origin("later"))))
    assert(withinBatch.map(_.earlierRecord.origin) == List(Some(Origin("earlier"))))

    // against the journal head the origin of the earlier writer is not at hand, see
    // `JournalFork.Record.fromJournalHead`
    val againstHead = JournalFork.fromEvents(key, journalHead(seqNr = 1, offset = 1), List(later))
    assert(againstHead.map(_.earlierRecord.origin) == List(none))
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
       val actual = stateT.run(initial)
       actual shouldEqual (expected, false).pure[Try]
     }
+
+    test(s"detect a journal fork appended in a later batch, $suffix") {
+      val key = Key("id", topic0)
+      val event0 = eventOf(SeqNr.unsafe(1), Offset.unsafe(1))
```
mr-git, line 2008: I wonder if test would become easier to read, if we would:
```scala
def eventOf(seqNr: Int, offset: Int)....
```
and here we would write:
```scala
val event0 = eventOf(seqNr = 1, offset = 1)
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+      // the stale write of the previous incarnation of the entity, landed after `event0` was already
+      // replicated: same `seqNr`, higher offset, different node and Cassandra timestamp
```
mr-git, line 2010: from where did you get `Cassandra` in this mix? the `timestamp` comes from `TopicRepliactor:122`
```suggestion
      // replicated: same `seqNr`, higher offset, different node and different timestamp
```

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+      val event1 = eventOf(SeqNr.unsafe(1), Offset.unsafe(2)).copy(origin = Some(origin1), timestamp = timestamp1)
```
mr-git, line 2011: I'd name events `event1` and `event2` - so their names manifest the offset they use

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+
+      val (state, _) = program.run(State.empty).get
+      val forks = state.forks
+
+      // `journal` clusters on `(seq_nr, timestamp)`, so both branches of the fork end up stored
+      state.journal.values.flatMap(_.keys).toSet shouldEqual
+        Set((SeqNr.min, timestamp0), (SeqNr.min, timestamp1))
+
+      forks.map(_.key) shouldEqual List(key)
+      forks.map(_.seqNr) shouldEqual List(SeqNr.min)
```
mr-git, line 2026:
```suggestion
      forks.map(_.seqNr) shouldEqual List(SeqNr.unsafe(1))
```
OR
in `event0` and `event1` definition use `SeqNr.min`

currently it requires hidden knowledge that `SeqNr.min == SeqNr.unsafe(1)`

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+      val event1 = eventOf(SeqNr.unsafe(1), Offset.unsafe(2)).copy(origin = Some(origin1), timestamp = timestamp1)
+
+      val program = journal
+        .append(key, Partition.min, Offset.unsafe(2), timestamp1, none, Nel.of(event0, event1))
+        .void
+
+      val (state, _) = program.run(State.empty).get
+      val forks = state.forks
+
+      forks.map(_.seqNr) shouldEqual List(SeqNr.min)
```
mr-git, line 2047:
```suggestion
      forks.map(_.seqNr) shouldEqual List(SeqNr.unsafe(1))
```
OR
extract `SeqNr.unsafe(1)` as value and use it here and in event constuctors

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+    }
+
+    test(s"detect a journal fork which regresses `seqNr` by more than one event, $suffix") {
+      val key = Key("id", topic0)
+      val event0 = eventOf(SeqNr.unsafe(1), Offset.unsafe(1))
+      val event1 = eventOf(SeqNr.unsafe(2), Offset.unsafe(2))
+      // the dead incarnation had both events in flight, so the live one replicated past the `seqNr`
+      // this one duplicates. Nothing here proves seqNr 1 is occupied - the head says 2 - so it is
+      // reported as suspected only, which is also what a legitimate out-of-order append looks like
+      val event2 = eventOf(SeqNr.unsafe(1), Offset.unsafe(3)).copy(origin = Some(origin1), timestamp = timestamp1)
```
mr-git, line 2061: I am not sure about "legitimate out-of-order append" part... how something so _legitimate_ could happen?

if previous `poll` was not committed and re-`poll` happened, then offset would match, but new offset means different publisher, which means that this is a true fork, doesn't it?

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+        _ <- journal.append(key, Partition.min, Offset.unsafe(3), timestamp1, none, Nel.of(event2))
+      } yield {}
+
+      val (state, _) = program.run(State.empty).get
+      val forks = state.forks
+
+      forks.map(_.seqNr) shouldEqual List(SeqNr.unsafe(1))
+      forks.map(_.duplicateProven) shouldEqual List(false)
+      // the fork's `seqNr` and the earlier one differ - the live incarnation got past what this
+      // event duplicates, so nothing here proves seqNr 1 is occupied
```
mr-git, line 2074: basically, it means that we have a "hidden" fork, which will never be tried during future recoveries - the client might have produced some important fork-ish side-effects, without negatively affecting the journal's future consistency?

eventual-cassandra/src/test/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandraTest.scala
```diff
@@ -1993,6 +2002,153 @@ class ReplicatedCassandraTest extends AnyFunSuite with Matchers {
...
+      val event0 = eventOf(SeqNr.unsafe(1), Offset.unsafe(1))
+      val event1 = eventOf(SeqNr.unsafe(2), Offset.unsafe(2))
+      val event2 = eventOf(SeqNr.unsafe(3), Offset.unsafe(4))
+      // a `seqNr` below the head again, reported as a regression like any other
+      val event3 =
+        eventOf(SeqNr.unsafe(2), Offset.unsafe(5)).copy(timestamp = timestamp1)
+
+      val program = for {
+        _ <- journal.append(key, Partition.min, Offset.unsafe(2), timestamp0, none, Nel.of(event0, event1))
+        _ <- journal.delete(key, Partition.min, Offset.unsafe(3), timestamp0, SeqNr.unsafe(2).toDeleteTo, none)
```
mr-git, line 2140: IMHO the following reads easier, if it is valid code:
```suggestion
        _ <- journal.delete(key, Partition.min, Offset.unsafe(3), timestamp0, DeleteTo(SeqNr.unsafe(2)), none)
```

eventual-cassandra/src/main/scala/com/evolution/kafka/journal/eventual/cassandra/ReplicatedCassandra.scala
rtar: I think we should, but it is not entirely trivial to do without breaking binary compatibility. Still thinking about it.

PR #971, opened 2026-08-12
add benchmarks for #957 change (new tests from #960 might need review or adjustments)

mr-git:
Related to #957 and #960.

Assuming that benchmark is valid, there is no improvement in cancellation handling.

```
Results on Apple M3 Max using Oracle JDK 17
Benchmark                                     Mode  Cnt     Score     Error  Units
ParFoldMap1Benchmark.originalPost11_0_0      thrpt    5  3419.803 ±  43.579  ops/s
ParFoldMap1Benchmark.originalPost11_0_0Plus  thrpt    5  3456.063 ±  42.177  ops/s
ParFoldMap1Benchmark.originalPre11_0_0       thrpt    5  3357.712 ± 155.907  ops/s
```

### DIFF
--- a/benchmark/src/test/scala/com/evolution/kafka/journal/replicator/ParFoldMap1Benchmark.scala
+++ b/benchmark/src/test/scala/com/evolution/kafka/journal/replicator/ParFoldMap1Benchmark.scala
@@ -1,0 +1,184 @@
+package com.evolution.kafka.journal.replicator
+
+import cats.data.{NonEmptyList, NonEmptyMap}
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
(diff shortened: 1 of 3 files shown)


mr-git:
this benchmark makes no sense:
* without long running (and sometimes failing) fibers the benchmark measures how many iterations we can call
* with long-running (`IO.sleep`) effects and with `.attempt` - we still count how fast the function executes
* the benchmarks do not measure how long the actual cancellation(s) take
* after semaphore and `parTraverse` fixes in Cats-Effect 3.7.1, there is no need for `.attempt` wrappers - we can afford to fail-fast on first error and `TopicReplicator` will attempt recovery sooner

Issue #961, opened 2026-08-01
Flaky CI: NoClassDefFoundError in KafkaConsumer shutdown hook (ConsistencySpec)


stasimus:
## What happens

`akkaTests`/`pekkoTests` CI jobs intermittently fail after all tests pass:

```
[info] Total number of tests run: 149
[info] All tests passed.
Error: Exception in thread "shutdownHook1" java.lang.NoClassDefFoundError: org/apache/kafka/common/utils/Utils$3
    at org.apache.kafka.common.utils.Utils.swallow(Utils.java:1048)
    at org.apache.kafka.clients.consumer.internals.ClassicKafkaConsumer.close(ClassicKafkaConsumer.java:1156)
    ...
    at com.evolutiongaming.skafka.consumer.Consumer$.$anonfun$1$$anonfun$1(Consumer.scala:263)
Caused by: java.lang.ClassNotFoundException: org.apache.kafka.common.utils.Utils$3
    at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
    ...
```

Seen in `ConsistencySpec` (akka/tests, pekko/tests) which forks its own JVM against real TestContainers Kafka/Cassandra.

## Root cause (best guess)

A JVM shutdown hook (registered to close a lingering `KafkaConsumer`) fires during forked-JVM teardown and lazily loads `Utils$3` (an anonymous class inside `Utils.swallow`) for the first time — racing against the classloader being torn down. `ClassicKafkaConsumer` is new in `kafka-clients` 4.3.x (replacing the old monolithic `KafkaConsumer` internals), so this looks tied to the recent `kafka-clients` 4.3.1 upgrade (#901, #945).

All actual tests pass before this happens — it only fails the build because the shutdown hook throws on its own thread, giving the process a nonzero exit code.

## Impact

- Failed 3/3 times on PR #934 (`fix/923-migrate-schema-swallows-errors`), which doesn't touch Kafka consumer code at all.
- Not consistently seen on master's recent runs, suggesting it's timing/race-dependent rather than deterministic.

## Suggested next steps

- Check if a newer `kafka-clients`/`skafka` version fixes this shutdown-hook path.
- Consider not relying on a JVM shutdown hook to close the consumer in the affected test helper, or explicitly pre-touching `Utils.swallow`'s code path before shutdown.

mr-git:
I tried to get the issue locally... there are two more, possibly related manifestations:
[...]

PR #988, opened 2026-08-28
Pekko integration tests - 3x speed up, big maintenance work

migesok:
Main points:
- "pekkoTests/testFull" reduced from 6 minutes to 2 minutes on M1
- fixed this issue by using test containers on available ports and not closing them explicitly on JVM exit (see SharedItEnv for details): https://github.com/evolution-gaming/kafka-journal/issues/961

Details:
- removed F[_] ceremonies from integration test code - use IO directly
- moved integration test classes into a separate package to simplify logging config
- renamed some integration tests for better readability
- removed unused parts, refactored integration test init code
- use consumer pool instead of creating a new consumer on each read in integration tests (the biggest test time win comes from this)
- reworked shared int test services (Cassandra, Kafka, replicator) to start on random available ports but without explicit stopping in a custom shutdown hook - this hardens the tests against potential env issues (occupied ports) and avoids the sbt 2 classloader close issue for forked test modules
- moved per-int-spec configuration overrides into config files named after the spec classes for better readability
- removed actor system usage from int tests where it wasn't needed
- fixed race in JournalPerfSpec - the test case code wasn't waiting for the test data prepare code

## Summary by CodeRabbit

* **Tests**
  * Consolidated integration testing around shared Kafka and Cassandra services.
  * Expanded coverage for journals, replication, settings, performance, JSON serialization, and Pekko plugins.
  * Added centralized configuration, resource cleanup, and suite lifecycle management.
* **Refactor**
  * Standardized integration application execution using a consistent runtime.
  * Removed obsolete test suites, utilities, serializers, and standalone configuration files.
  * Improved test logging and centralized service connectivity settings.


pekko/tests/src/test/scala/com/evolution/kafka/journal/app/package.scala
```diff
@@ -0,0 +1,7 @@
+package com.evolution.kafka.journal
+
+// migesok:
+// I moved all ad-hoc apps not really doing anything for integration tests into this package.
+// Not sure why they are needed, probably can be removed.
```
mr-git, line 5: they look like an examples on how to use `kafka-journal`, or can be used as a "seed" for special needs, like data recovery or similar

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/pekko/PekkoJournalPluginPerfSpec.scala
```diff
@@ -0,0 +1,11 @@
+package com.evolution.kafka.journal.it.pekko
+
+import com.evolution.kafka.journal.it.SharedItEnv
+import org.apache.pekko.persistence.journal.JournalPerfSpec
+
+class PekkoJournalPluginPerfSpec extends JournalPerfSpec(
```
mr-git, line 6: please, let's use full words
```suggestion
class PekkoJournalPluginPerformanceSpec extends JournalPerfSpec(
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItInstances.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.util.Try
+
+/**
+ * Combines typeclass instances used in integration tests to reduce code duplication.
+ *
+ * It is assumed that the integration tests use [[IO]] directly.
+ *
+ * Only "lightweight" instances are put here - the ones which do not require special disposal logic.
+ */
+private[it] object ItInstances {
```
mr-git, line 18: please, let's use full words

```suggestion
private[it] object IntegrationTestInstances {
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItOrigin.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.evolution.kafka.journal.Origin
+import org.scalatest.Suite
+
+/**
+ * Provides [[Origin]] value with the test suite name for integration tests.
+ *
+ * Could be used to mark the journal data written by integration tests. Which could help to
+ * distinguish which data belongs to which test suite.
+ */
+private[it] trait ItOrigin { this: Suite =>
```
mr-git, line 12: please, let's use full words

```suggestion
private[it] trait IntegrationTestOrigin { this: Suite =>
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalCirceIntSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import cats.effect.IO
+import cats.implicits.*
+import com.evolution.kafka.journal.circe
+import com.evolution.kafka.journal.conversions.{KafkaRead, KafkaWrite}
+import com.evolution.kafka.journal.eventual.EventualRead
+import com.evolution.kafka.journal.{Event, SeqNr}
+import io.circe.Json
+
+class JournalCirceIntSpec extends JournalIntSpec[Json] {
```
mr-git, line 11: please, let's use full words

```suggestion
class JournalCirceIntegrationSpec extends JournalIntSpec[Json] {
```

mr-git:
I haven't finished the review, but I dislike that git doesn't show the "linked" changes - there are just too many changes for it (and me) to grasp:
* `F` -> `IO` could be done in separate PR
* file renames in dedicated PR
* setup refactoring (essence of this PR) could stay here

Please use full names instead of acronyms, like:
* `It` -> `IntegrationTest`
* `Int` -> `Integration` or `IntegrationTest`
* `Perf` -> `Performance`

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalIntSpec.scala
```diff
@@ -0,0 +1,296 @@
...
+  "Journal" - {
+    journalImplsToTest.foreach { testJournalImpl =>
+      s"for impl ${ testJournalImpl.describe }" - {
+
+        val key = Key.random[IO]("journal")
+
+        // msokolov:
+        // Journal release logic here is broken, it was broken before my changes, and I'm not sure how to fix it.
+        // release is called in the last test case, and it is not properly released if the last test case is
+        // skipped.
```
mr-git, line 78: can we run explicitly 4 different and correctly named unit-tests, where each would call the function `journalsResource` with corresponding parameters and each could handle their own resource

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalPerfSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  private val eventualCassandra = allocateSuiteScoped(eventualCassandraResource)
+  private val journalProducer = allocateSuiteScoped(journalProducerResource)
+  private val journalConsumerPool = allocateSuiteScoped(journalConsumerPoolResource)
+
+  private val journalImplsToTest: Vector[TestJournalImpl] = Vector(
+    new TestJournalImpl(
+      eventualJournalImplName = "empty",
+      eventualJournalImpl = EventualJournal.empty[IO],
+      avgReadOpTimeAcceptanceThreshold = 2.seconds,
```
mr-git, line 31: we need a comment, on how this constant was decided

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalPerfSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+    ),
+    new TestJournalImpl(
+      eventualJournalImplName = "cassandra",
+      eventualJournalImpl = eventualCassandra,
+      avgReadOpTimeAcceptanceThreshold = 1.seconds,
```
mr-git, line 36: we need a comment, on how this constant was decided

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalSuite.scala
```diff
@@ -0,0 +1,130 @@
...
+import com.evolution.kafka.journal.eventual.{EventualJournal, EventualRead}
+import com.evolutiongaming.catshelper.{Log, LogOf}
+import org.scalatest.Suite
+import org.scalatest.matchers.should.Matchers
+
+import java.time.Instant
+
+/**
+ * Base trait for Kafka-Journal integration tests not using Pekko and working with journal
+ * interfaces below the Pekko persistence plugin.
```
mr-git, line 20: what "below" means? "using" or "inside"?

```suggestion
 * interfaces using same API as Pekko persistence plugin uses.
```

mr-git:
Please split this PR into dedicated smaller PRs - one for each change:
* move files around without changing their content besides package (so that git diff and I can track them)
* rename files where the renaming is due
* switch from `F` to `IO`
* the essence of this PR

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
```diff
@@ -0,0 +1,260 @@
...
+ * Obtain an instance using [[SharedItEnv.require()]]. If the call succeeds, it means that these
+ * services have been started:
+ *   - Kafka in testcontainers
+ *   - Cassandra in testcontainers
+ *   - Kafka-Journal replicator in the same JVM
+ *
+ * Kafka and Cassandra containers are bound to random available ports - use [[SharedItEnv]] instance
+ * values and methods to configure Kafka-Journal components connectivity.
+ *
+ * It is expected that integration tests using this component are run in a forked JVM, as it relies
```
mr-git, line 35: > It is expected that ..

what has to be done to make sure that it is the case?

if somebody will change some (which?) setting(s) in (I assume) `build.sbt` - will this fail with some clear reason?

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
```diff
@@ -0,0 +1,260 @@
...
+   *   test suite class
+   */
+  def loadSuiteConfig(suiteClass: Class[?]): JournalItConfig = {
+    JournalItConfig.parseUnsafe(configLoader.loadUnsafe(
+      testOverride = ConfigFactory.parseResources(s"${ suiteClass.getSimpleName }.conf"),
+    ))
+  }
+
+  /**
+   * [[loadSuiteConfig(java.lang.Class)]] but with a type param
```
mr-git, line 71: is this a syntax sugar?

```suggestion
   * Syntax sugar for [[loadSuiteConfig(java.lang.Class)]], so it looks more functional and is easier to use
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
```diff
@@ -0,0 +1,260 @@
...
+  )
+  private case object Kafka extends TestContainerDef[KafkaContainer](
+    image = KafkaDockerImage,
+    makeContainer = new KafkaContainer(_),
+    getConnString = _.getBootstrapServers,
+  )
+
+  private lazy val memoizedStartEnvResult: Either[Throwable, SharedItEnv] = {
+    Try {
+      val (env, _) = runEnvResource.allocated.unsafeRunSync()
```
mr-git, line 103: shall we document why we do not care about `release` here?

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
```diff
@@ -0,0 +1,260 @@
...
+      configLoader = configLoader,
+    )
+  }
+
+  private def replicatorResource(config: ReplicatorConfig): ResourceIO[Unit] = {
+    logResourceStart(name = "replicator") {
+      for {
+        metrics <- Replicator.Metrics.make[IO](CollectorRegistry.empty[IO], "fake-client-id")
+        hostName <- HostName.of[IO]().toResource
+        replicatorRunIo <- Replicator.make[IO](config, cassandraClusterOf, hostName, metrics.some)
```
mr-git, line 151: why do we have to create the `metrics` on `empty` registry? can we?:
```suggestion
        hostName <- HostName.of[IO]().toResource
        replicatorRunIo <- Replicator.make[IO](config, cassandraClusterOf, hostName, none)
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  private abstract class TestContainerDef[C <: GenericContainer[C]](
+    image: DockerImageName,
+    makeContainer: DockerImageName => C,
+    getConnString: C => String,
+  ) { this: Product =>
+    private def name: String = productPrefix
+
+    private def logLifeCycleEvent(msg: String): Unit = {
+      log.info(s"[Test-container $name] $msg")
```
mr-git, line 182: is it `TestContainer` (as the name of this abstract clas) or `testcontainer` (as a name of library)?

```suggestion
      log.info(s"[TestContainer $name] $msg")
```

OR

```suggestion
      log.info(s"[testcontainer $name] $msg")
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SuiteScopedResources.scala
```diff
@@ -0,0 +1,94 @@
...
+ *       ...
+ *     }
+ *
+ *     // test code using heavyComponent1 and heavyComponent2 goes here
+ *     ...
+ *   }
+ * }}}
+ *
+ * In case allocation fails for one of the resources, all previously created resources are
+ * deallocated and the failure is propagated to the suite constructor.
```
mr-git, line 34: in Cats-Effect lingo we use `release`, not `deallocate`:

```suggestion
 * released and the failure is propagated to the suite constructor.
```

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SuiteScopedResources.scala
```diff
@@ -0,0 +1,94 @@
...
+ * Scala 2.13 it might result in deadlocks.
+ */
+private[it] trait SuiteScopedResources extends BeforeAndAfterAll { this: Suite =>
+  import ItInstances.*
+  import SuiteScopedResources.*
+
+  private[this] val releaseIoDeque = new ConcurrentLinkedDeque[IO[Unit]]
+
+  /**
+   * Allocates a test-suite-scoped resource with automatic destruction.
```
mr-git, line 46:
```suggestion
   * Allocates a test-suite-scoped resource with automatic release.
```

mr-git (changes requested):

mr-git:
Overall: I am failing to follow up with improvements interleaved with heavy refactoring, I am *NOT* going to compare thousands of lines of unit-tests to check if something was [accidentally] changed!

mr-git:
Also, when this PR will get split - please include the same activities on akka parts too!

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/pekko/PekkoJournalPluginPerfSpec.scala
rtar: I think PerfSpec comes from Akka/Pekko vocabulary, here, so I would actually keep the shortening in this case.

pekko/tests/src/test/scala/com/evolution/kafka/journal/app/AppendReplicateApp.scala
```diff
@@ -1,77 +1,77 @@
...
+    Resource.make(IO {
+      ActorSystem("AppendReplicateApp", config)
+    }) { system =>
+      IO.fromFuture(IO {
+        system.terminate()
+      }).void
+    }
   }
 
-  private def runF[F[_]: Async: Parallel: FromGFuture: MeasureDuration: FromAttempt: FromTry: ToTry: Fail](
```
rtar, line 42: I think one of the reasons `runF` was used here is to have a place, where Kafka Journal authors could see / try the ergonomics of Kafka Journal in Tagless Final environment.

Kafka Journal is often used in libraries, to hide complexity of it, so it is kind of a main use case.

By removing the pain of TF here, we move this pain to the consumer of the library because they lose the only full working example of how the journal could be constructed in TF world.

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalIntSpec.scala
```diff
@@ -0,0 +1,296 @@
...
+import org.scalatest.Succeeded
+import org.scalatest.freespec.AsyncFreeSpec
+import play.api.libs.json.Json
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import scala.concurrent.duration.*
+
+abstract class JournalIntSpec[A] extends AsyncFreeSpec with JournalSuite {
+  import JournalIntSpec.*
```
rtar: Should we move it to top of the page?

pekko/tests/src/test/scala/com/evolution/kafka/journal/app/package.scala
migesok: I'm uneasy about code placement for those - may be move them to another submodule?

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/pekko/PekkoJournalPluginPerfSpec.scala
migesok: Yes, my naming is consistent with Akka/Pekko vocab, it is deliberate here.

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
migesok: no this is a "test container" as a thing, a docker container for tests

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItInstances.scala
migesok: I don't think this improves the code anyhow in this case

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalCirceIntSpec.scala
migesok: I don't think this improves the code anyhow in this case

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItInstances.scala
mr-git: Then why in some cases we use `It`, but in other `Int`?

Let's move all integration tests to dedicated module `integration-test` and then we will have all classes in it locally scoped.

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalCirceIntSpec.scala
mr-git: in dedicated module, the infix will not be required at all

pekko/tests/src/test/scala/com/evolution/kafka/journal/app/AppendReplicateApp.scala
migesok: I will preserve the example code while moving it into a different dedicated module

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
migesok: applied, thanks!

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/SharedItEnv.scala
migesok: it is described in the SharedItEnv doc but I added a comment line here anyway

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/JournalCirceIntSpec.scala
migesok: I like to keep integration tests having a special suffix in their name. Even with a dedicated module (which it is already) you might have some int-test things unit tested in the same module as well.

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItOrigin.scala
migesok: already discussed, rejected

pekko/tests/src/test/scala/com/evolution/kafka/journal/it/ItInstances.scala
migesok: already discussed, rejected

pekko/tests/src/test/scala/com/evolution/kafka/journal/app/package.scala
migesok: decided to move the apps into a `-demo` module

migesok:
Will be split into separate PRs as decided.

PR #989, opened 2026-08-28
Set up single status check for all CI action sub-jobs

migesok:
This is needed for master branch protection rule.

## Summary by CodeRabbit

* **Chores**
  * Added an aggregate continuous integration status check.
  * Branch protection can now rely on a single status that reflects the overall test result.


mr-git:
can we make it as part of shared workflows?

PR #991, opened 2026-08-31
Move demo apps into a separate module

migesok:
Also:
* remove actor system dependency from the demo apps - it wasn't really needed

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated demo module with sample applications for appending, replicating, and reading events.
  - Added demo configuration for Kafka and Cassandra connectivity.
  - Added asynchronous console logging configuration for demo applications.

- **Refactor**
  - Simplified demo application startup and configuration handling.
  - Removed obsolete test-only demo applications and configurations.
  - Consolidated demo application packaging and resource management.


demo/src/main/resources/AppendReplicateApp.conf
```diff
@@ -0,0 +1,38 @@
+# AppendReplicateApp settings
+
+# msokolov: I have no idea why these specific overrides are needed, could be removed?
```
rtar, line 3: I have investigated just for sake of it. I will add the comments near specific overrides.

demo/src/main/resources/AppendReplicateApp.conf
```diff
@@ -0,0 +1,38 @@
...
+
+evolutiongaming.kafka-journal {
+
+  persistence.journal {
+    kafka {
+      bootstrap-servers = [
```
rtar, line 9: These come from the `scripts/*` configuration. It was, likely, meant to run the Kafka cluster locally and then run `AppendReplicateApp` manually on them. It is possible that the scripts do not work anymore and need update themselves, but it might be worth to address it in a separate PR to avoid a scope creep.

stasimus (approved):

build.sbt
```diff
@@ -345,6 +346,28 @@ lazy val ScalaTestIO = project
...
+ */
+lazy val demo = project
+  .in(file("demo"))
+  .settings(name := "kafka-journal-demo")
+  .settings(commonSettings)
+  .settings(
+    Seq(
+      publish / skip := true,
+    ),
+  )
```
mr-git, line 360: like in other places:

```suggestion
  .settings(commonSettings)
  .settings(publish / skip := true)
```

demo/src/main/resources/AppendReplicateApp.conf
mr-git: please add such questions either:
* as an open issue with link to file and/or location
* as a comment in PR

why we have to keep in repo the issues, which are not clearly marked as TODO?!

```suggestion
# TODO why these specific overrides are needed, could they be removed?
```

why ask rhetorical question, when you can easily remove them and attempt to start the app locally?

demo/src/main/resources/AppendReplicateApp.conf
```diff
@@ -0,0 +1,38 @@
...
+    }
+  }
+
+  replicator {
+    kafka = ${evolutiongaming.kafka-journal.persistence.journal.kafka}
+    cassandra = ${evolutiongaming.kafka-journal.persistence.journal.cassandra}
+
+    topic-discovery-interval = 1s
+  }
+}
```
mr-git, line 38:
```suggestion
}

```

PR #992, opened 2026-09-02
Integration test refactoring

migesok:
* remove F[_] ceremonies and use IO[_] directly
* remove actor system usage in integration tests which didn't need it

## Summary by CodeRabbit

* **Tests**
  * Standardized integration and performance tests around a consistent execution model.
  * Centralized shared test setup for database, messaging, logging, serialization, and runtime support.
  * Improved journal test handling for optional record-timestamp overrides.
  * Simplified configuration loading and container startup for integration tests.

* **Refactor**
  * Removed redundant test utilities and duplicated setup logic.
  * Simplified test resource management and dependency initialization.


akka/tests/src/test/scala/com/evolution/kafka/journal/circe/JournalCirceIntSpec.scala
```diff
@@ -1,7 +1,7 @@
 package com.evolution.kafka.journal.circe
 
 import cats.effect.IO
-import cats.syntax.all.*
+import cats.implicits.*
```
mr-git: why all these `cats.syntax.all.*` -> `cats.implicits.*` changes?

migesok: Isn't it the preferred way to import cats syntax now?

akka/tests/src/test/scala/com/evolution/kafka/journal/IntegrationSuite.scala
```diff
@@ -38,118 +39,99 @@ object IntegrationSuite {
...
-            new HostConfig()
-              .withPortBindings(new PortBinding(
-                Ports.Binding.bindPort(exposeStaticPort),
-                new ExposedPort(exposeStaticPort),
-              )),
-          )
-          ()
-        }
+  ): IO[C] = {
+    IO {
```
mr-git, line 47: why we do not need `.blocking` marker any more?

akka/tests/src/test/scala/com/evolution/kafka/journal/JournalSuite.scala
```diff
@@ -81,78 +71,46 @@ trait JournalSuite extends ActorSuite with Matchers { self: Suite =>
...
-
-      def size[A](
-        implicit
-        kafkaRead: KafkaRead[F, A],
-        eventualRead: EventualRead[F, A],
-      ): F[Long] = journal.read().length
-    }
+      kafkaRead: KafkaRead[IO, A],
+      eventualRead: EventualRead[IO, A],
+    ): IO[Long] = journal.read().length
```
mr-git: it is nicer, when function's body dangles under definition:

```suggestion
    def pointer: IO[Option[SeqNr]] =
      journal.pointer

    def delete(to: DeleteTo): IO[Option[PartitionOffset]] =
      journal.delete(to)

    def purge: IO[Option[PartitionOffset]] =
      journal.purge

    def size[A](
      implicit
      kafkaRead: KafkaRead[IO, A],
      eventualRead: EventualRead[IO, A],
    ): IO[Long] =
      journal.read().length
```

akka/tests/src/test/scala/com/evolution/kafka/journal/SettingsIntSpec.scala
```diff
@@ -117,27 +100,27 @@ class SettingsIntSpec extends AsyncWordSpec with BeforeAndAfterAll with Matchers
 
         for {
           a <- get(setting.key)
-          _ <- Sync[F].delay { a shouldEqual None }
+          _ <- IO { a shouldEqual None }
```
mr-git, line 103: frankly, here and in all other places, we do not need the `IO { ... }` wrapper for checks, do we?

mr-git (commented):
same comments apply for pekko changes too

akka/tests/src/test/scala/com/evolution/kafka/journal/circe/JournalCirceIntSpec.scala
mr-git: it is a noise in diff, it is better to make such changes in dedicated PR, instead of changing _some_ files in random PRs.

as for `Isn't it the preferred way`, it is complicated - `implicits.*` import a different set of implicits, on some edge cases they are not equivalent. 

https://typelevel.org/cats/imports.html, https://typelevel.org/cats-effect/docs/concepts - use `.syntax.all.*` form.

https://github.com/typelevel/cats/issues/4332#issuecomment-1297685027

https://github.com/typelevel/cats/releases/tag/v2.2.0 - check the first sentence of `Migration`

https://github.com/typelevel/cats-effect/issues/4462#issuecomment-3164918199 - there is comment, but here Daniel mentions that `implicits.*` import might, and historically has, managed to negatively affect resolution of performance of implicit resolutions.

mr-git: let's not pollute this PR with unneeded changes!

akka/tests/src/test/scala/com/evolution/kafka/journal/IntegrationSuite.scala
migesok: because this operation is not blocking

akka/tests/src/test/scala/com/evolution/kafka/journal/circe/JournalCirceIntSpec.scala
migesok: removed

akka/tests/src/test/scala/com/evolution/kafka/journal/SettingsIntSpec.scala
migesok: I don't think it hurts either here.

akka/tests/src/test/scala/com/evolution/kafka/journal/JournalSuite.scala
migesok: fixed

akka/tests/src/test/scala/com/evolution/kafka/journal/circe/JournalCirceIntSpec.scala
mr-git: frankly, my comment wasn't about single case, but about general approach - it adds noise in about half of affected files, but never mind - we can align them in other PR

mr-git (approved):

